PR #5, opened 2022-02-07
Replace requires with custom error messages

x676f64:


ed-iv:
I too am an enjoyer of custom errors

martinloesethjensen:
Cool didn't know about custom errors in solidity

PR #26, opened 2022-06-13
RMRKEquippable refactor with custom errors

stevyhacker:
RMRKEquippableMock was over the contract size limit so I replaced all string errors in RMRKEquippable with custom errors.
Tests are updated too and now the RMRKEquippableMock contract uses  **700** bytes less to deploy.

Please be careful when reviewing if I refactored all require conditions to IFs properly as they are inverted now.


steven2308 (approved):
Awesome, thanks!

PR #27, opened 2022-06-14
Refactors all common from MultiResource and Equippables

steven2308:
This includes events, all internal variables except for resources and all functions which do not require access to resources, since structs are different in both implementations.


steven2308:
This is the new layout
6  Inheritance Alternative

PR #28, opened 2022-06-14
Update Nesting contracts with custom errors

stevyhacker:
I have continued with custom error updates, mostly in the NestingAbstract contract.
Also, I commented out hardhat/console.sol as I don't think we should commit it enabled and it's not used anywhere at the moment anyway.


steven2308 (approved):
Looks great!
Regarding the hardhat console, we use it from time to time so it's just easier to keep it there for now. But I guess commented is good.
Thanks!

steven2308:
|  RMRKEquippableMock            ·     23.503  ·       -0.886  │

PR #35, opened 2022-06-24
Add base storage tests

stevyhacker:
I have added more tests for the RMRKBaseStorage contract.
Everything seems fine except the checkIsEquippableMulti function which doesn't work as expected, I tried to debug it but couldn't find any logical reason why it returns an array out of bounds error.

### DIFF
--- a/test/baseStorage.ts
+++ b/test/baseStorage.ts
@@ -1,94 +1,125 @@
-// import { ethers } from 'hardhat';
-// import { expect } from 'chai';
-// import { RMRKBaseStorageMock } from '../typechain';
-// import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers';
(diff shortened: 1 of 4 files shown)


test/baseStorage.ts
```diff
@@ -1,94 +1,125 @@
...
+
+    it('cannot multi check equippables with unequal input arrays', async function () {
+      const id = 6;
+      const id2 = 7;
+      await expect(
+        testBase.checkIsEquippableMulti([id, id2], [addrs[1].address]),
+      ).to.be.revertedWith('RMRKMismatchedInputArrayLength()');
+    });
+
+    it('check is equippable multi should return true', async function () {
```
stevyhacker, line 111: It's this one @CicadaNCR

PR #41, opened 2022-07-28
Implement test fixtures

stevyhacker:
Fixtures allow Hardhat to not repeat the same operations and load the correct testing state rather than redoing all the setup on each test in the beforeEach loop. I implemented fixtures for a part of the testing codebase and here are the results:

Before fixtures  -  92fe24b551c432e21c083b401222196433b8a9b8
________________________________________________________
Executed in  **158.30 secs**
   usr time  148.80 secs    0.06 millis  148.80 secs
   sys time    9.02 secs    1.19 millis    9.02 secs

After implementing fixtures for half the tests - c05b7c4b6c6390a5853f37feeccba460aca97e4c
________________________________________________________
Executed in   **71.10 secs**
   usr time   68.83 secs    0.07 millis   68.83 secs
   sys time    4.14 secs    1.21 millis    4.13 secs
   
I don't like how some of these code changes look, but let me know if you see how it can be improved.
I think the most useful fixtures are for the equippable stuff as there is a lot of setup for each contract before testing besides the deployments themselves.


stevyhacker:
Added more test fixtures and merged latest changes from /dev

PR #47, opened 2022-08-04
feat: modify `_equipments` mapping path

Ayuilos:
Seems like this modification does not influence anything 😉


steven2308 (commented):
Related resources (which I guess means equipped) from different collections can have the same Id. This would not allow it.

steven2308:
Example:
You equip a weapon with resource Id 1. You want to equip a helmet with resource Id 1, you're screwed.  
IMO you're trying to solve a "problem" which is not really a problem if you simply use another base.

Ayuilos:
Why I'm screwed? When I excute `equip`, the `childIndex` and `childResourceId` could separate it.

steven2308:
I thought you were using the childResourceId, not the parent's. So my counter example does not apply.
It's still different from the spec, [here](https://github.com/rmrk-team/rmrk-spec/blob/master/standards/rmrk2.0.0/interactions/equip.md#baseslot-explained) it explains why it uses both base and slot to reference equipments.
I'm not sure what consequences could there be by not using the base in the mapping.

PR #49, opened 2022-08-10
Fixes inconsistency on parent nft when burning a nested child,

steven2308:
also adds options for owners to remove references to children regardless of the state of the child.

### DIFF
--- a/contracts/RMRK/RMRKEquippable.sol
+++ b/contracts/RMRK/RMRKEquippable.sol
@@ -278,6 +278,16 @@ contract RMRKEquippable is IRMRKEquippable, MultiResourceAbstract {
         }
     }
 
+    function forgetEquipment(
+        uint256 tokenId,
+        uint64 resourceId,
+        uint64 slotId
+    ) external onlyOwnerOrApproved(tokenId) {
+        address targetBaseAddress = _baseAddresses[resourceId];
+        delete _equipments[tokenId][targetBaseAddress][slotId];
+        // FIXME: Shall it emit an event?
+    }
+
     //Gate for equippable array in here by check of slotPartDefinition to slotPartId
     function composeEquippables(
         uint tokenId,
--- a/contracts/RMRK/RMRKNesting.sol
+++ b/contracts/RMRK/RMRKNesting.sol
@@ -28,6 +28,7 @@ error RMRKUnnestForNonexistentToken();
 error RMRKUnnestForNonNftParent();
 error RMRKUnnestFromWrongChild();
 error RMRKNotApprovedOrOwnerOrChild();
+error RMRKChildIsNotOrphan();
 
 contract RMRKNesting is ERC721, IRMRKNesting {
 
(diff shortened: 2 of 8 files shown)


steven2308:
Going back to the initial problems:

1. Rejected children still exist on the child Contract -> Solved with child.unnestSelfOrphan (manual call)
2. Removed children still exist on the child Contract -> Solved with child.unnestSelfOrphan (manual call)
3. When burning a nested token, it's nor removed from the parent -> Solved with parent.forgetChild (automatic call, and can also be manual). If equippables are in use, we also need a manual call to parentEquip.forgetEquipment.

Ayuilos:
I think the `forgetChild` is better to become a extension of `RMRKNesting` token, the reason is:
- You can't guarantee every contract will follow this, the `forgetChild` are more like a listener interface in design, the worse case is parent still record the burned broken child, and even **the worst case** is the bad implementer tells you it has a child, but in fact the child recorded owner isn't the bad implementer contract.
- If you can't not avoid to do validation before you use it, then I think it's better to just left the core functions to keep clean and tiny, others all better become extensions.

contracts/RMRK/RMRKEquippable.sol
```diff
@@ -278,6 +278,16 @@ contract RMRKEquippable is IRMRKEquippable, MultiResourceAbstract {
...
     }
 
+    function forgetEquipment(
+        uint256 tokenId,
+        uint64 resourceId,
+        uint64 slotId
+    ) external onlyOwnerOrApproved(tokenId) {
+        address targetBaseAddress = _baseAddresses[resourceId];
+        delete _equipments[tokenId][targetBaseAddress][slotId];
+        // FIXME: Shall it emit an event?
```
stevyhacker, line 288: Emitting an event here makes sense

contracts/RMRK/RMRKNesting.sol
```diff
@@ -386,10 +392,25 @@ contract RMRKNesting is ERC721, IRMRKNesting {
...
     //Child-scoped interaction
+    // FIXME: Should emit event
     function _unnestSelf(uint256 tokenId, uint256 indexOnParent) internal virtual {
-      // A malicious contract which is parent to this token, could unnest any children
         RMRKOwner memory owner = _RMRKOwners[tokenId];
+        _transferToRootOwner(tokenId, owner);
+        IRMRKNesting(owner.ownerAddress).unnestChild(owner.tokenId, tokenId, indexOnParent);
+    }
+
+    function _unnestOrphanSelf(uint256 tokenId) internal virtual {
```
stevyhacker, line 402: I'm not a fan of this naming but I am not sure what would sound better

CicadaNCR:
> I think the forgetChild is better to become a extension of RMRKNesting...

Just for some context, as a general design principle we're modeling these after the ERC721 spec and openzeppelin implementations, which means:

1. We provide the user with an interface contract -- This is the only 'true' test of whether or not a token is RMRK-standard compliant. Similar to ERC721, it is not possible to enforce specific implementations in any way, shape or form.

2. We provide an example implementation of the RMRK standard, which will enforce everything we believe necessary to enforce in the context of the its own storage; because we can't enforce a given implementation (and we all agree on this) the best we can do is ensure that:

- All contract calls in the context of a given implementer-as-rote behave appropriately even when calling potentially malicious contracts
- A given implementer-as-rote will not be able to brick itself in the course of normal operations.

These two conditions should be satisfied by the default implementation. So while having a dedicated function to 'recover' from a broken state isn't ideal, it's still a path towards necessary functionality. Thus, this should be a core element of the contract.

steven2308:
This will be discarded, we went for an alternative of making everything parent scoped.

PR #55, opened 2022-09-02
Size reduce to make equippables standalone implementation compile.

steven2308:


contracts/RMRK/RMRKEquippable.sol
```diff
@@ -204,28 +170,8 @@ contract RMRKEquippable is RMRKNesting, IRMRKEquippable {
         return getResource(resourceId);
     }
 
-    function getFullResources(uint256 tokenId)
```
stevyhacker, line 207: Removing these additional view functions really keeps the main contract lean, I like it 👍

stevyhacker (approved):

contracts/RMRK/RMRKEquippable.sol
steven2308: Yeah! We should do that on the EIP too.

PR #57, opened 2022-09-12
Security changes

steven2308:
These are mainly related to visibility. All variables where made private instead of internal, and most of external methods were made public so custom implementations can use them.

### DIFF
--- a/contracts/RMRK/library/RMRKLib.sol
+++ b/contracts/RMRK/library/RMRKLib.sol
@@ -32,7 +32,7 @@ library RMRKLib {
         returns (uint256, bool)
     {
         uint256 length = A.length;
-        for (uint256 i = 0; i < length; ) {
+        for (uint256 i; i < length; ) {
             if (A[i] == a) {
                 return (i, true);
             }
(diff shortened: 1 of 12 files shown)


contracts/RMRK/library/RMRKLib.sol
```diff
@@ -32,7 +32,7 @@ library RMRKLib {
         returns (uint256, bool)
     {
         uint256 length = A.length;
-        for (uint256 i = 0; i < length; ) {
+        for (uint256 i; i < length; ) {
```
stevyhacker, line 35: Is this change more gas effective?

ThunderDeliverer: I think there shouldn't be any difference once compiled

steven2308: I didn't actually test it. I'm following this:
https://github.com/transmissions11/solcurity

ThunderDeliverer: You could also review everything using Mythril: https://github.com/ConsenSys/mythril should give you a feeling about improvement pre- and post-update

stevyhacker (approved):

PR #61, opened 2022-09-16
Adds additional level on approvals

steven2308:
The problem:
When transfering a token with nested children, the children approvals are not updated, it would be very expensive to do so and malicious children may prevent it. This affects the regular approval and the approval for resources with the following vulnerability:
1. I approve a child token to a secondary address of mine
2. Sell the parent
3. I could do changes on resources for child (approved for resources)
4. If the child is later unnested, I could transfer it using the approved address.

Solution:
Instead of keeping track of approvals per token (token->approved), we add an extra level which is the approver. New mapping is token->approver->approved.
When checking if an address is approved, we check against the token Id and current owner, so if the root owner changes, permissions are invalidated.
WARNING: If the root owner eventually goes back to the original approver, these permissions would be valid again.

### DIFF
--- a/contracts/RMRK/equippable/RMRKEquippable.sol
+++ b/contracts/RMRK/equippable/RMRKEquippable.sol
@@ -57,8 +57,10 @@ contract RMRKEquippable is RMRKNesting, IRMRKEquippable {
 
     // ------------------- RESOURCE APPROVALS --------------
 
-    // Mapping from token ID to approved address for resources
-    mapping(uint256 => address) private _tokenApprovalsForResources;
+    // Mapping from token ID to approver address to approved address for resources
+    // The approver is necessary so approvals are invalidated for nested children on transfer
+    // WARNING: If a child NFT returns the original root owner, old permissions would be active again
+    mapping(uint256 => mapping(address => address)) private _tokenApprovalsForResources;
 
     // Mapping from owner to operator approvals for resources
     mapping(address => mapping(address => bool))
@@ -325,7 +327,7 @@ contract RMRKEquippable is RMRKNesting, IRMRKEquippable {
         returns (address)
     {
         _requireMinted(tokenId);
-        return _tokenApprovalsForResources[tokenId];
+        return _tokenApprovalsForResources[tokenId][ownerOf(tokenId)];
     }
 
     function setApprovalForAllForResources(address operator, bool approved)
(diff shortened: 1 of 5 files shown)


contracts/RMRK/equippable/RMRKEquippable.sol
```diff
@@ -373,16 +375,17 @@ contract RMRKEquippable is RMRKNesting, IRMRKEquippable {
         internal
         virtual
     {
-        _tokenApprovalsForResources[tokenId] = to;
-        emit ApprovalForResources(ownerOf(tokenId), to, tokenId);
+        address owner = ownerOf(tokenId);
+        _tokenApprovalsForResources[tokenId][owner] = to;
+        emit ApprovalForResources(owner, to, tokenId);
```
ThunderDeliverer, line 380: Am I correct to assume that access control has to be handled by the implementer?

contracts/RMRK/equippable/RMRKEquippable.sol
```diff
@@ -373,16 +375,17 @@ contract RMRKEquippable is RMRKNesting, IRMRKEquippable {
...
     }
 
-    function _cleanApprovals(address owner, uint256 tokenId)
+    function _cleanApprovals(uint256 tokenId)
```
ThunderDeliverer, line 383: We could also clear all of the approvals on the token using this method, not only the approvals set by current owner 🤔

stevyhacker: Clearing all approvals might be safer for the user as then we want have the original old permissions back.

> // WARNING: If a child NFT returns the original root owner, old permissions would be active again

contracts/RMRK/equippable/RMRKEquippable.sol
steven2308: For most operations we already provide access control to the token owner or approved only.

contracts/RMRK/equippable/RMRKEquippable.sol
steven2308: That would be ideal, but you cannot clear a mapping. :smiling_face_with_tear:

steven2308: One thing we could do to improve this warning situation is to also clear approvals on incoming NFTs. It will change nothing for nested NFTs though. It would just help if I approved something, nested into another NFT, it left my wallet and it came back unnested. :/

PR #65, opened 2022-09-20
Remove factory code from EVM repo

stevyhacker:
Just removing factory-related code from the EVM repo, as that is now kept on the https://github.com/rmrk-team/evm-marketplace


steven2308 (approved):
Awesome, thanks!

PR #66, opened 2022-09-20
Add Royalties IERC2981 implementation

stevyhacker:
Adds an abstract contract with Royalties implemented with IERC2981.

Updating royalty recipient is added as an unimplemented method as it requires some kind of access control and I didn't want to include that in the interface as it clashed with OwnableLock on other contracts.

Updating royalty percentage is not added as Bruno disliked the option of the collection owner being able to change royalty to any percent after the collection deployment so this field can only be set once in the collection deployment.


steven2308 (approved):
Looks great, I agree on not including the gated function to set royalties.

PR #67, opened 2022-09-20
A simple script to generate abis jsons from implementation contracts

Yuripetusko:
A simple script to generate abis jsons from implementation contracts


CicadaNCR:
Seems fine on its face -- any reason in particular you're writing them to `/artifacts/abis/implementations/` instead of just loading them directly?

Yuripetusko:
> Seems fine on its face -- any reason in particular you're writing them to `/artifacts/abis/implementations/` instead of just loading them directly?

Directly like where? I'm open to suggestions of a better directory to save them to. For now I just copy and paste the whole folder to the Singular codebase, so it's easy for me to do it if they are all written in the same dir

CicadaNCR:
> 

That makes sense, I'll merge since it's just a utility script anyway. By default you can find the generated ABIs their *.sol folder under the contracts directory, Nesting for example:

`artifacts/contracts/implementations/RMRKNestingImpl.sol/RMRKNestingImpl.json`

I realize walking the tree could be a PITA, so as a utility script that runs automagic on some other action, is all good.

Yuripetusko:
Yeah I cannot simply see these files in viewer,. only when I inspection sol files with code is when I see json files, so this seemed like an easy option

PR #71, opened 2022-09-26
Changes on RMRKBaseStorage

steven2308:
Moves struct to interface
Removes symbol
Adds metadataURI
Adds getType and getMetadataURI to interface.

### DIFF
--- a/contracts/RMRK/base/IRMRKBaseStorage.sol
+++ b/contracts/RMRK/base/IRMRKBaseStorage.sol
@@ -57,6 +57,24 @@ interface IRMRKBaseStorage is IERC165 {
         string metadataURI; //n bytes 32+
     }
 
+    //TODO: Doc this struct, put JSON intake format in comments here
+    struct IntakeStruct {
+        uint64 partId;
+        Part part;
+    }
+
+    /**
+     * @dev Returns URI of the metadata of associated collection
+     * @return string base contract metadata URI
+     */
+    function getMetadataURI() external view returns (string memory);
+
+    /**
+     * @dev Returns type of data of associated base
+     * @return string data type
+     */
+    function getType() external view returns (string memory);
+
     /**
      * @dev Returns true if the part at partId is equippable by targetAddress.
      *
--- a/contracts/RMRK/base/RMRKBaseStorage.sol
+++ b/contracts/RMRK/base/RMRKBaseStorage.sol
@@ -30,19 +30,12 @@ contract RMRKBaseStorage is IRMRKBaseStorage {
 
     uint64[] private _partIds;
 
-    string private _symbol;
+    string private _metadataURI;
     string private _type;
 
-    //TODO: Move to interface
-    //TODO: Doc this struct, put JSON intake format in comments here
-    struct IntakeStruct {
-        uint64 partId;
-        Part part;
-    }
-
-    constructor(string memory symbol_, string memory type__) {
-        _symbol = symbol_;
-        _type = type__;
+    constructor(string memory metadataURI, string memory type_) {
+        _metadataURI = metadataURI;
+        _type = type_;
     }
 
     /**
@@ -53,7 +46,7 @@ contract RMRKBaseStorage is IRMRKBaseStorage {
         _;
     }
 
-    function _onlySlot(uint64 partId) internal view {
+    function _onlySlot(uint64 partId) private view {
         ItemType itemType = _parts[partId].itemType;
         if (itemType == ItemType.None) revert RMRKPartDoesNotExist();
         if (itemType == ItemType.Fixed) revert RMRKPartIsNotSlot();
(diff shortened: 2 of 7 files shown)


steven2308:
@ThunderDeliverer  I had to do a change here so I also changed the getters to avoid the weird `type__` thingy on the issue you reported. What do you think?

contracts/RMRK/base/IRMRKBaseStorage.sol
```diff
@@ -57,6 +57,24 @@ interface IRMRKBaseStorage is IERC165 {
         string metadataURI; //n bytes 32+
     }
 
+    //TODO: Doc this struct, put JSON intake format in comments here
```
ThunderDeliverer, line 60: Just FYI this is done in the examples documentation. I'll add it once that is merged

contracts/RMRK/base/RMRKBaseStorage.sol
```diff
@@ -30,19 +30,12 @@ contract RMRKBaseStorage is IRMRKBaseStorage {
...
-    //TODO: Doc this struct, put JSON intake format in comments here
-    struct IntakeStruct {
-        uint64 partId;
-        Part part;
-    }
-
-    constructor(string memory symbol_, string memory type__) {
-        _symbol = symbol_;
-        _type = type__;
+    constructor(string memory metadataURI, string memory type_) {
```
ThunderDeliverer, line 36: Is there a reason that `metadataURI` doesn't have a trailing underscore like most of the input arguments?
Also what do you think about changing the capitalisation to `metadataUri`, like we do for `resourceId`?
Both of the changes kind of unify the way we do it in most of the functions.

contracts/RMRK/base/RMRKBaseStorage.sol
```diff
@@ -53,7 +46,7 @@ contract RMRKBaseStorage is IRMRKBaseStorage {
         _;
     }
 
-    function _onlySlot(uint64 partId) internal view {
+    function _onlySlot(uint64 partId) private view {
```
ThunderDeliverer, line 49: Shouldn't this be available in the contracts that inherit this one?

contracts/RMRK/base/RMRKBaseStorage.sol
```diff
@@ -77,15 +70,15 @@ contract RMRKBaseStorage is IRMRKBaseStorage {
...
     }
 
     /**
      * @dev Returns type of data of associated base
      * @return string data type
      */
-    function type_() public view returns (string memory) {
+    function getType() external view returns (string memory) {
         return _type;
     }
```
ThunderDeliverer, line 83: We could just change the visibility of the `_type` variable to `public` and have the autogenerated getter. Wouldn't this reduce the file size?

contracts/implementations/RMRKBaseStorageImpl.sol
```diff
@@ -16,8 +16,8 @@ import "../RMRK/access/OwnableLock.sol";
  */
 
 contract RMRKBaseStorageImpl is OwnableLock, RMRKBaseStorage {
-    constructor(string memory symbol_, string memory type__)
-        RMRKBaseStorage(symbol_, type__)
+    constructor(string memory metadataURI, string memory type_)
```
ThunderDeliverer, line 19: Same question about the possibility of changing the `metadataURI` to ` metadataUri_` as above.

contracts/mocks/RMRKBaseStorageMock.sol
```diff
@@ -5,8 +5,8 @@ pragma solidity ^0.8.15;
 import "../RMRK/base/RMRKBaseStorage.sol";
 
 contract RMRKBaseStorageMock is RMRKBaseStorage {
-    constructor(string memory symbol_, string memory type__)
-        RMRKBaseStorage(symbol_, type__)
+    constructor(string memory metadataURI, string memory type_)
```
ThunderDeliverer, line 8: Just flagging this `metadataURI` as well.

ThunderDeliverer (commented):
Just added a few questions and comments. :)

contracts/RMRK/base/RMRKBaseStorage.sol
steven2308: It does the opposite, increases file size

contracts/RMRK/base/RMRKBaseStorage.sol
steven2308: We only add the trailing underscore if needed (to avoid shadowing a method usually). Here there's no need because the method is named "get.."  As for the capitalization I'm indifferent but on ERC721 Metadata extension they use `tokenURI` so uppercase URI feels more standard.

contracts/RMRK/base/RMRKBaseStorage.sol
steven2308: Do you see any reason? 
I think the modifier should be enough, no need to call this directly. This is actually put into a separate function just to save some contract size.

contracts/RMRK/base/RMRKBaseStorage.sol
ThunderDeliverer: Ok, I'm cool with this, people interacting our contracts won't need to worry about it and as long as we know why we do it the way we do, it's fine 👌

contracts/RMRK/base/RMRKBaseStorage.sol
ThunderDeliverer: I could see some use cases where people would want to use it in their own business logic, but on the other hand, if they need it, they can implement it themselves

contracts/RMRK/base/RMRKBaseStorage.sol
ThunderDeliverer: Ok, then the point is moot 😄

ThunderDeliverer (approved):
LGTM 🚀

PR #72, opened 2022-09-26
Checks for inheritance loop, up to 100 levels of nesting.

steven2308:
If you create an inheritance loop, ownership checks will always run out of gas due to an infinite loop. 
We avoid this at least up to 10 levels.


steven2308:
@CicadaNCR  this is something we were missing together with preventing self nesting.
To avoid it being bypassed by adding more nesting levels, we simply raise another exception after we go through 100 levels without finding a root owner.
The alternative would be to prevent so many levels on minting/nesting but it's very complex because you would have to check up an down for cases like this:
1. A->B->C->D->E->.....->F
2. P->Q->-R->-T->U->...
3. Nest P into F.

Yuri and I found this a good compromise. What do you think?

stevyhacker (approved):

PR #85, opened 2022-10-10
Update RMRKRoyalties wtih RoyaltyDetails struct

stevyhacker:
I updated the RMRKRoyalties with the RoyaltyDetails struct which I use to avoid the limit of too many params in the factories. 
Also, I added public getters for royalty recipient and royalty percentage.


steven2308:
This is only needed for the factory contracts so we'll keep it clean and not do it here.
We'll only add the `RMRKRoyalties.getRoyaltyRecipient` method.

PR #86, opened 2022-10-13
Reentrancy security pass

ThunderDeliverer:
I flagged some of the instances where there is potential for malicious actor to abuse the external calls as well as some other instances that might warrant a discussion.

### DIFF
--- a/contracts/RMRK/multiresource/RMRKMultiResource.sol
+++ b/contracts/RMRK/multiresource/RMRKMultiResource.sol
@@ -344,7 +344,7 @@ contract RMRKMultiResource is
         _beforeTokenTransfer(owner, address(0), tokenId);
 
         // Clear approvals
-        _approve(address(0), tokenId);
+        _approve(address(0), tokenId); // I don't get why we are setting approval for 0x0 if we are also deleting it (assigning it to the same address)
         _approveForResources(address(0), tokenId);
 
         _balances[owner] -= 1;
--- a/contracts/implementations/RMRKEquippableImpl.sol
+++ b/contracts/implementations/RMRKEquippableImpl.sol
@@ -76,7 +76,7 @@ contract RMRKEquippableImpl is
         uint256 mintPriceRequired = numToMint * _pricePerMint;
         if (mintPriceRequired != msg.value) revert RMRKMintUnderpriced();
 
-        uint256 nextToken = _totalSupply + 1;
+        uint256 nextToken = _totalSupply + 1; // Doesn't this cause gaps in the IDs? So for example, if this would be the first mint ever on a collection, it would set the initial ID of the token to 1 instead of 0.
         unchecked {
             _totalSupply += numToMint;
         }
--- a/contracts/RMRK/nesting/RMRKNesting.sol
+++ b/contracts/RMRK/nesting/RMRKNesting.sol
@@ -287,6 +287,8 @@ contract RMRKNesting is Context, IERC165, IERC721, IRMRKNesting, RMRKCore {
         _sendToNFT(tokenId, destinationId, from, to);
     }
 
+    // This function calls an address passed as parameter when burning a child. This could potentailly be used as reentrancy vector.
+    // We might want to potentially import the reentrancy guard and invoke it here.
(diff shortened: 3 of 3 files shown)


contracts/RMRK/multiresource/RMRKMultiResource.sol
```diff
@@ -344,7 +344,7 @@ contract RMRKMultiResource is
         _beforeTokenTransfer(owner, address(0), tokenId);
 
         // Clear approvals
-        _approve(address(0), tokenId);
+        _approve(address(0), tokenId); // I don't get why we are setting approval for 0x0 if we are also deleting it (assigning it to the same address)
```
steven2308, line 347: On 721 the just delete it like this: `delete _tokenApprovals[tokenId];` . 
We cannot do something that simply though, because our approvals are a nested mapping, not a simple one:
`mapping(uint256 => mapping(address => address)) private _tokenApprovals;`
The reason is on top of the definition:
```
// Mapping from token ID to approver address to approved address
// The approver is necessary so approvals are invalidated for nested children on transfer
// WARNING: If a child NFT returns the original root owner, old permissions would be active again
```

contracts/implementations/RMRKEquippableImpl.sol
```diff
@@ -76,7 +76,7 @@ contract RMRKEquippableImpl is
         uint256 mintPriceRequired = numToMint * _pricePerMint;
         if (mintPriceRequired != msg.value) revert RMRKMintUnderpriced();
 
-        uint256 nextToken = _totalSupply + 1;
+        uint256 nextToken = _totalSupply + 1; // Doesn't this cause gaps in the IDs? So for example, if this would be the first mint ever on a collection, it would set the initial ID of the token to 1 instead of 0.
```
steven2308, line 79: We don't want people to use tokenId 0, it will mess up in some validations. Maybe we should be more explicit about this or even enforce it somewhere (maybe inside `_mint`?)

ThunderDeliverer: We should definitely emphasise this somewhere. Maybe examples first and then in the EVM docs as well (once we have them)? 🤔

steven2308: Actually we need discourage 0 in general for ids. On resources, base parts, equippable groups. Will take note.

ThunderDeliverer: Yeah, I'm pretty sure the majority of engineers will try to start with ID 0. We need to enforce this in some manner.

stevyhacker: I agree with this point, starting from id = 1 should be pointed out both in the docs and the examples, especially because the rest of the  NFT industry uses token ids beginning from 0.

contracts/RMRK/multiresource/RMRKMultiResource.sol
steven2308: I answered thinking of transferring.
For burn we could avoid it we assume tokens are not reminted, but I don't think it hurts to have this here.

contracts/implementations/RMRKEquippableImpl.sol
ThunderDeliverer: Could we give a special status to ID 0? Like a representation of the collection? So NFT with ID 0 wouldn't be a regular NFT, but rather a 'cover' NFT of the collection? 🤔

steven2308: The thing with Id 0 is that we use a lot of mappings where Ids are either keys or in values. When querying for something if the value is 0 we assume there's no entry. So having a custom logic for 0 is probably too complex.
For 'cover' or collection data we already have collection meta data (extension).

ThunderDeliverer:
I think we can close this PR without merging. Anyone opposed? @steven2308 @stevyhacker

steven2308:
Agree, let's close it.

ThunderDeliverer:
As we have discussed the possible implications and added the guards on other chain, this PR doesn't bring any more value to the repository

PR #89, opened 2022-10-17
Nested hooks

steven2308:
Adds nesting hooks and some questions about missing restrictions.

### DIFF
--- a/contracts/RMRK/nesting/RMRKNesting.sol
+++ b/contracts/RMRK/nesting/RMRKNesting.sol
@@ -245,11 +245,13 @@ contract RMRKNesting is Context, IERC165, IERC721, IRMRKNesting, RMRKCore {
         address to,
         uint256 tokenId
     ) internal virtual {
-        (address immediateOwner, , ) = rmrkOwnerOf(tokenId);
+        // TODO: Shall we require parentId to be 0? Otherwise should a be a _nestTransfer call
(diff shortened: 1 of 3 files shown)


contracts/RMRK/nesting/RMRKNesting.sol
```diff
@@ -929,6 +956,22 @@ contract RMRKNesting is Context, IERC165, IERC721, IRMRKNesting, RMRKCore {
...
+        uint256 tokenId
+    ) internal virtual {}
+
+    function _afterNestedTokenTransfer(
+        address from,
+        address to,
+        uint256 fromTokenId,
+        uint256 toTokenId,
+        uint256 tokenId
+    ) internal virtual {}
```
ThunderDeliverer, line 973: I like the content of the hooks. Should include everything that we need 👌

contracts/RMRK/nesting/RMRKNesting.sol
```diff
@@ -245,18 +245,21 @@ contract RMRKNesting is Context, IERC165, IERC721, IRMRKNesting, RMRKCore {
         address to,
         uint256 tokenId
     ) internal virtual {
-        (address immediateOwner, , ) = rmrkOwnerOf(tokenId);
+        // TODO: Shall we require parentId to be 0? Otherwise should a be a _nestTransfer call
```
ThunderDeliverer, line 248: I think so. Let's enforce the desired behaviour.

PR #90, opened 2022-10-17
Forbids setting 0 as id for tokens, resources, parts, and equippable groups

steven2308:


contracts/RMRK/multiresource/RMRKMultiResource.sol
```diff
@@ -23,9 +23,10 @@ error ERC721TokenAlreadyMinted();
 error ERC721TransferFromIncorrectOwner();
 error ERC721TransferToNonReceiverImplementer();
 error ERC721TransferToTheZeroAddress();
-error RMRKNotApprovedForResourcesOrOwner();
 error RMRKApprovalForResourcesToCurrentOwner();
 error RMRKApproveForResourcesCallerIsNotOwnerNorApprovedForAll();
+error RMRKNotApprovedForResourcesOrOwner();
+error RMRKTokenIdZeroForbidden();
```
ThunderDeliverer, line 29:
```suggestion
error RMRKIdZeroForbidden();
```

Could we just use the `RMRKIdZeroForbidden` error name, so that the same error is triggered for ID 0 across the board? We have a lot of custom errors. Having a general error telling users that ID 0 is not supported in our standards might be useful.

For example: If the user gets the `RMRKTokenIdZeroForbidden` they might be surprised to find out that slots are not allowed to use ID 0 as well.

stevyhacker: I agree with Jan here, it will be more clear to the users if we make this change.

steven2308: That was the plan, but due to our inheritance flow I get a conflict because it's define on AbstractMultiResource and RMRKNesting :( 
I don't see another workaround. Do you guys see one?

steven2308: Because of the MR Only and Nesting Only versions. If we leave it just in one, the other one will miss it.

steven2308: Only thing I can think of now is to have a library of errors which everything imports, I think it might solve the conflict issues but that would be a big change.

steven2308: I could not find a solution, will merge and take not of it.

PR #93, opened 2022-10-19
Adds hooks on multiresource management methods.

steven2308:


rdovgan (approved):

stevyhacker (approved):
I am not sure if we need both before and after or just after for resource rejection but it looks good overall and because it doesn't cost much it might be good to have it.

steven2308:
Better have both and not need them than need both and not have them 😅
This one did not even add size to the contracts, probably the compiler strips it out if it's empty.

PR #94, opened 2022-10-21
Extract errors to a library

ThunderDeliverer:
Errors were extractred to a library smart contract. The contract option was picked over the library simply because deployment in such manner should conserve gas. The errors library smart contracts was then imported in the smart contracts and the inheritance tree respected, so that each instance only gets it inherited once.

### DIFF
--- a/contracts/RMRK/library/RMRKErrors.sol
+++ b/contracts/RMRK/library/RMRKErrors.sol
@@ -20,15 +20,15 @@ error ERC721ApprovedQueryForNonexistentToken();
 error ERC721ApproveToCaller();
 /// Attempting to use an invalid token ID
 error ERC721InvalidTokenId();
-/// Attempting to mint to 0x0 addressß
+/// Attempting to mint to 0x0 address
 error ERC721MintToTheZeroAddress();
 /// Attempting to manage a token without being its owner or approved by the owner
 error ERC721NotApprovedOrOwner();
 /// Attempting to mint an already minted token
 error ERC721TokenAlreadyMinted();
-/// Attempting to transfer the token form and address that is not the owner
+/// Attempting to transfer the token from an address that is not the owner
 error ERC721TransferFromIncorrectOwner();
-/// Attempting to transfer to an address that is unable to receive the token
+/// Attempting to safe transfer to an address that is unable to receive the token
 error ERC721TransferToNonReceiverImplementer();
 /// Attempting to transfer the token to a 0x0 address
 error ERC721TransferToTheZeroAddress();
@@ -36,11 +36,9 @@ error ERC721TransferToTheZeroAddress();
 error RMRKApprovalForResourcesToCurrentOwner();
 /// Attempting to grant approval of resources without being the caller or approved for all
 error RMRKApproveForResourcesCallerIsNotOwnerNorApprovedForAll();
-/// Attempting to grant resource approval to self
-error RMRKApproveForResourcesToCaller();
 /// Attempting to incorrectly configue a Base item
 error RMRKBadConfig();
-/// Attempting to set the priorities with an array of length that doesn't match the resource length
+/// Attempting to set the priorities with an array of length that doesn't match the length of active resources array
 error RMRKBadPriorityListLength();
 /// Attempting to add a resource entry without a `Part`
 error RMRKBaseRequiredForParts();
@@ -50,7 +48,7 @@ error RMRKCannotTransferSoulbound();
 error RMRKChildAlreadyExists();
 /// Attempting to interact with a child, using index that is higher than the number of children
 error RMRKChildIndexOutOfRange();
-/// Attempting to equip a `Part` that does not allow equipping to the desired address
+/// Attempting to equip a `Part` with a child not approved by the base
 error RMRKEquippableEquipNotAllowedByBase();
 /// Attempting to use ID 0, which is not supported
 /// @dev The ID 0 in RMRK suite is reserved for empty values. Guarding against its use ensures the expected operation
@@ -59,9 +57,9 @@ error RMRKIdZeroForbidden();
 error RMRKIndexOutOfRange();
 /// Attempting to reclaim a child that can't be reclaimed
 error RMRKInvalidChildReclaim();
-/// Attempting to interact with an end-user accoutn when the contract account is expected
+/// Attempting to interact with an end-user account when the contract account is expected
 error RMRKIsNotContract();
-/// Attempting to interact with a contract that had its operation paused
+/// Attempting to interact with a contract that had its operation locked
 error RMRKLocked();
 /// Attempting to add a pending child after the number of pending children has reached the limit (default limit is 128)
 error RMRKMaxPendingChildrenReached();
@@ -86,14 +84,16 @@ error RMRKNestingTransferToSelf();
 error RMRKNoResourceMatchingId();
 /// Attempting to manage a resource without owning it or having been granted permission by the owner to do so
 error RMRKNotApprovedForResourcesOrOwner();
-/// Attempting to interact with a token or resource without being its owner or having been granted permission by the
+/// Attempting to interact with a token without being its owner or having been granted permission by the
 ///  owner to do so
+/// @dev When a token is nested, only the direct owner (NFT parent) can mange it. In that case, approved addresses are
+///  not allowed to manage it, in order to ensure the expected behaviour
 error RMRKNotApprovedOrDirectOwner();
-/// Attempting to compose a resource wihtout an address
+/// Attempting to compose a resource wihtout having an associated Base
 error RMRKNotComposableResource();
 /// Attempting to unequip an item that isn't equipped
 error RMRKNotEquipped();
-/// Attempting to interact with a token or resource without being its owner
+/// Attempting to interact with a management function without being the smart contract's owner
 error RMRKNotOwner();
 /// Attempting to interact with a function without being the owner or contributor of the collection
 error RMRKNotOwnerOrContributor();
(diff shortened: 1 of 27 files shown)


steven2308:
This is much cleaner than I expected, really nice. I just removed the contract definition since i's not really necessary. May I merge it?

ThunderDeliverer:
Please don't merge until I push the commit containing the error documentation, I'll be done in the evening and we can merge then (if I don't do it now, I'll have to adapt the documentation plan 😅 )

ThunderDeliverer:
This is now ready for review 👌

ThunderDeliverer:
@steven2308 I have no objection to your unrelated commits 🙌 🚀

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
+/// Attempting to grant the token to 0x0 address
 error ERC721AddressZeroIsNotaValidOwner();
+/// Attempting to grant approval to the current owner of the token
 error ERC721ApprovalToCurrentOwner();
+/// Attempting to grant approval when not being owner or approved for all should not be permitted
 error ERC721ApproveCallerIsNotOwnerNorApprovedForAll();
+/// Attempting to get approvals for a token owned by 0x0 (considered non-existent)
 error ERC721ApprovedQueryForNonexistentToken();
+/// Attempting to grant approval to self
 error ERC721ApproveToCaller();
```
steven2308, line 20: Let's merge this into ERC721ApprovalToCurrentOwner.

ThunderDeliverer: Could this not be emitted if the caller was approved for all and tried to call approval on specific token for self?

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
 error RMRKNestingTransferToNonRMRKNestingImplementer();
+/// Attempting to nest the token into itself
 error RMRKNestingTransferToSelf();
+/// Attempting to interact with a resource that can not be found
 error RMRKNoResourceMatchingId();
+/// Attempting to manage a resource without owning it or having been granted permission by the owner to do so
 error RMRKNotApprovedForResourcesOrOwner();
+/// Attempting to interact with a token or resource without being its owner or having been granted permission by the
+///  owner to do so
 error RMRKNotApprovedOrDirectOwner();
```
steven2308, line 91: Let's remove " resource" and add this:
/// @dev when a token is nested, only the direct owner (NFT parent) can mange it, approved addresses are not allowed in that case to ensure consistency.

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
+/// Attempting to compose a resource wihtout an address
```
steven2308, line 92: /// Attempting to compose a resource without a base associated

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
 error RMRKNotComposableResource();
+/// Attempting to unequip an item that isn't equipped
 error RMRKNotEquipped();
+/// Attempting to interact with a token or resource without being its owner
```
steven2308, line 96: This is actually the contract owner, shall we rename it?
It guards operations as locking the contract, transfer ownership and withdrawing the value from the contract.

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
 error RMRKPartAlreadyExists();
+/// Attempting to use a `Part` that doesn't exist
 error RMRKPartDoesNotExist();
+/// Attempting to use a `Part` that is `Fixed` when `Slot` kind of `Part` should be used
 error RMRKPartIsNotSlot();
+/// Attempting to interact with a pending child using an index greater than the size of pending array
 error RMRKPendingChildIndexOutOfRange();
+/// Attempting to add a resource using an ID that has already been used
 error RMRKResourceAlreadyExists();
+/// Attempting to equip an item into a slot that already has an item euipped
```
steven2308, line 114: Typo equipped

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
 error RMRKSlotAlreadyUsed();
+/// Attempting to equip an item into a resource that is not a `Slot`
```
steven2308, line 116: /// Attempting to equip an item into a slot that the target resource does not implement

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
 error RMRKTargetResourceCannotReceiveSlot();
+/// Attempting to equip a child into a slot that doesn't support the tokens from child's collection
```
steven2308, line 118: /// Attempting to equip a child into a slot and parent that the child does not support.

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
 error RMRKTokenCannotBeEquippedWithResourceIntoSlot();
+/// Attempting to compose a NFT of a token without active resources
 error RMRKTokenDoesNotHaveActiveResource();
+/// Attempting to determine a priority of tokens resources wihtout them being defined
```
steven2308, line 122: /// Attempting to determine the resource with the top priority on a token with no resources added.

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
 error RMRKTokenHasNoResources();
-error RMRKZeroLengthIdsPassed();
+/// Attempting not to pass an empth array of equippable addresses when adding or setting the equippable addresses
```
steven2308, line 124: typo: empty

steven2308 (changes requested):
In some cases the name of the error might be misleading, if that's the case let's discuss better options.

stevyhacker:
This looks much cleaner and I like that the errors have a single place for documentation too!

contracts/RMRK/library/RMRKErrors.sol
steven2308: What you describe is the case on MultiResource so I guess we need to keep it.
`if (owner == operator) revert ERC721ApproveToCaller();` 
On Nesting and all inheriting implementations, we have an extra level on approvals which is always set to the sender, so there we could use the one I suggested. But it might look inconsistent :thinking: 
`if (_msgSender() == operator) revert ERC721ApproveToCaller();`

Both checks are on `setApprovalForAll`

ThunderDeliverer: I'd leave these two separate for now and we can address them if we find a good and consistent solution. At least we have a centralised errors file where we can do it :D

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
 error ERC721InvalidTokenId();
+/// Attempting to mint to 0x0 addressß
 error ERC721MintToTheZeroAddress();
+/// Attempting to manage a token without being its owner or approved by the owner
 error ERC721NotApprovedOrOwner();
+/// Attempting to mint an already minted token
 error ERC721TokenAlreadyMinted();
+/// Attempting to transfer the token form and address that is not the owner
 error ERC721TransferFromIncorrectOwner();
+/// Attempting to transfer to an address that is unable to receive the token
```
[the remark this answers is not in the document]
ThunderDeliverer: Sure, makes it more verbose

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
 error ERC721TransferToNonReceiverImplementer();
+/// Attempting to transfer the token to a 0x0 address
 error ERC721TransferToTheZeroAddress();
+/// Attempting to grant approval of resources to their current owner
 error RMRKApprovalForResourcesToCurrentOwner();
+/// Attempting to grant approval of resources without being the caller or approved for all
 error RMRKApproveForResourcesCallerIsNotOwnerNorApprovedForAll();
+/// Attempting to grant resource approval to self
 error RMRKApproveForResourcesToCaller();
```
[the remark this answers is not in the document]
ThunderDeliverer: Done :D

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
+/// Attempting to incorrectly configue a Base item
 error RMRKBadConfig();
+/// Attempting to set the priorities with an array of length that doesn't match the resource length
```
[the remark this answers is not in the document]
ThunderDeliverer: How about this?

contracts/RMRK/library/RMRKErrors.sol
ThunderDeliverer: Hm, I'll fix the error description for now, but we can create a separate task to rename owner of the contract into something like `Issuer`, `Creator`, `Maintainer`, ...

contracts/RMRK/library/RMRKErrors.sol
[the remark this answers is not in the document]
steven2308: What do you mean?
I'm just pointing at a missing "s" :stuck_out_tongue:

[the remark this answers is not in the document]
ThunderDeliverer: Hahaha, I was replying to comments in the VS code while doing the changes, so the response might make more sense now, that I pushed them 😂

contracts/RMRK/library/RMRKErrors.sol
```diff
@@ -2,59 +2,124 @@
...
 error RMRKBadPriorityListLength();
+/// Attempting to add a resource entry without a `Part`
```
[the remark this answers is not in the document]
steven2308: @ThunderDeliverer  only this one is missing.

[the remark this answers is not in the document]
ThunderDeliverer:
```suggestion
/// Attempting to add a resource entry with `Part`s, without setting the `Base` address
```

steven2308 (approved):
Great change in general!

PR #95, opened 2022-10-24
Implements NestTransfer event.

steven2308:


steven2308:
As in the nest transfer hooks, the from is the immediate owner, not the root.

stevyhacker (approved):

PR #99, opened 2022-10-31
Implements bounded recursive burning.

steven2308:


test/behavior/nesting.ts
```diff
@@ -416,6 +417,49 @@ async function shouldBehaveLikeNesting(
...
+      // 0 is not enough
+      await expect(parent.connect(tokenOwner)['burn(uint256,uint256)'](parentId, 0))
+        .to.be.revertedWithCustomError(parent, 'RMRKMaxRecursiveBurnsReached')
+        .withArgs(child.address, childId);
+      // 1 is not enough
+      await expect(parent.connect(tokenOwner)['burn(uint256,uint256)'](parentId, 1))
+        .to.be.revertedWithCustomError(parent, 'RMRKMaxRecursiveBurnsReached')
+        .withArgs(child.address, grandChild1);
+      // 2 is not enough
+      await expect(parent.connect(tokenOwner)['burn(uint256,uint256)'](parentId, 2))
```
stevyhacker, line 448: It would look a bit nicer if typesafe calls were used here but it's not a big deal

stevyhacker (approved):

test/behavior/nesting.ts
steven2308: Yep, added task. This is something I'd like to. We need to figure out how to do it from behavior suites.

PR #100, opened 2022-10-31
Splits all management operations into a gated and an internal version

steven2308:
This allows custom implementers to by pass gating under certain conditions as
- Minting
- Factories
- Game Contracts

### DIFF
--- a/contracts/RMRK/equippable/RMRKEquippable.sol
+++ b/contracts/RMRK/equippable/RMRKEquippable.sol
@@ -204,18 +204,18 @@ contract RMRKEquippable is
 
     // ------------------------------- EQUIPPING ------------------------------
 
-    function unnestChild(
+    function _unnestChild(
         uint256 tokenId,
         uint256 index,
         address to,
         bool isPending
-    ) public virtual override onlyApprovedOrOwner(tokenId) {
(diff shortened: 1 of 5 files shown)


contracts/RMRK/equippable/RMRKEquippable.sol
```diff
@@ -204,18 +204,18 @@ contract RMRKEquippable is
...
     // ------------------------------- EQUIPPING ------------------------------
 
-    function unnestChild(
+    function _unnestChild(
         uint256 tokenId,
         uint256 index,
         address to,
         bool isPending
-    ) public virtual override onlyApprovedOrOwner(tokenId) {
+    ) internal virtual override {
```
stevyhacker, line 212: So right now all gating is on the implementations?

stevyhacker (approved):

contracts/RMRK/equippable/RMRKEquippable.sol
steven2308: Only collection owner gating is on implementations. (Add resource, Add resource to token)
For token owner, all the core contracts now have both the public gated version which calls the internal one. (Accept resources and children, equip, etc).

PR #103, opened 2022-11-07
Prevents race condition on nesting

steven2308:
On acceptChild and unnestChild you now need to send childAddress and childId, in addition to the childIndex. 
Index is used to cheaply locate the child, while address and id are used to check it is the expected child we're dealing with.

Also renames parentTokenId -> parentId and childTokenId -> childId, for simplicity.


test/behavior/nesting.ts
```diff
@@ -248,37 +248,47 @@ async function shouldBehaveLikeNesting(
...
     });
 
     it('cannot accept not owned child', async function () {
       const nowOwner = addrs[3];
-      await expect(parent.connect(nowOwner).acceptChild(parentId, 0)).to.be.revertedWithCustomError(
-        parent,
-        'ERC721NotApprovedOrOwner',
-      );
+      await expect(
+        parent.connect(nowOwner).acceptChild(parentId, 0, child.address, childId),
```
ThunderDeliverer, line 274: Is `nowOwner` a typo or should it be called that?

test/behavior/nesting.ts
```diff
@@ -544,15 +586,17 @@ async function shouldBehaveLikeNesting(
       const toOwner = tokenOwner.address;
       const notOwner = addrs[3];
       await expect(
-        parent.connect(notOwner).unnestChild(parentId, 0, toOwner, false),
+        parent.connect(notOwner).unnestChild(parentId, toOwner, 0, child.address, childId, false),
       ).to.be.revertedWithCustomError(child, 'ERC721NotApprovedOrOwner');
     });
 
-    it('cannot unnest not existing child', async function () {
+    it('cannot unnest from not existing parent', async function () {
```
ThunderDeliverer, line 593: Can we add the correct example to check that a non-existent child can't be unnested as well?
It might be hidden by the GitHub UI, so it might already exist

test/behavior/nesting.ts
```diff
@@ -643,15 +719,17 @@ async function shouldBehaveLikeNesting(
       const toOwner = tokenOwner.address;
       const notOwner = addrs[3];
       await expect(
-        parent.connect(notOwner).unnestChild(parentId, 0, toOwner, true),
+        parent.connect(notOwner).unnestChild(parentId, toOwner, 0, child.address, childId, true),
       ).to.be.revertedWithCustomError(child, 'ERC721NotApprovedOrOwner');
     });
 
-    it('cannot unnest not existing child', async function () {
+    it('cannot unnest from not existing parent', async function () {
```
ThunderDeliverer, line 726: Same question as above. Should we add another example to validate that attempting to unnest a child with invalid index gets reverted as well?

ThunderDeliverer (changes requested):
I see no issues here, the NatSpec comments are most of the requested changes.

I like the changes to the variable names, makes them easier to read (and hopefully to use 😅 )

contracts/RMRK/equippable/RMRKEquippable.sol
```diff
@@ -311,25 +311,34 @@ contract RMRKEquippable is
      * @notice Used to unnest a given child.
      * @dev The function doesn't contain a check validating that `to` is not a contract. To ensure that a token is not
      *  transferred to an incompatible smart contract, custom validation has to be added when using this function.
-     * @param tokenId ID of the token we are unnesting a child from
-     * @param index Index of a token we are unnesting in the array it belongs to (can be either active array or pending
-     *  array)
-     * @param to End user address to unnest the token to
-     * @param isPending Specifies whether the child being unnested is in the pending array (`true`) or in the active
-     *  array (`false`)
+     * @param tokenId is the tokenId of the parent token to unnest from.
```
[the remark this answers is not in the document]
steven2308: Great to know, will check, thanks!

test/behavior/nesting.ts
steven2308: The non existent child is now covered on `cannot accept child if address or id do not match` and `cannot accept children for non existing index` 
This was just a bad name from the beginning

test/behavior/nesting.ts
steven2308: It was a typo, thanks

steven2308:
@ThunderDeliverer thanks for the suggestions, it should be all covered now.

test/behavior/nesting.ts
ThunderDeliverer: Great, then I see no additions needed 👌

ThunderDeliverer (approved):
🚢 🚀 🎆

PR #105, opened 2022-11-08
Gets rid of Extended resource and saves some space in the process

steven2308:
The trade off was to remove functions to access equippable info by resource:
    - `getSlotPartIds(uint64 resourceId)`
    - `getFixedPartIds(uint64 resourceId)`
    - `getBaseAddressOfResource(uint64 resourceId)`
    
    In favor of having `getExtendedResource(uint256 tokenId, uint64 resourceId)` return all information.


contracts/RMRK/equippable/RMRKEquippable.sol
```diff
@@ -196,32 +196,26 @@ contract RMRKEquippable is
...
     ) internal virtual {
-        uint64 id = resource.id;
-        _addResourceEntry(id, resource.metadataURI);
+        _addResourceEntry(id, metadataURI);
 
         if (
-            resource.baseAddress == address(0) &&
+            baseAddress == address(0) &&
             (fixedPartIds.length != 0 || slotPartIds.length != 0)
         ) revert RMRKBaseRequiredForParts();
```
stevyhacker, line 215: Having a Base is a requirement for each Equippable?

stevyhacker: So having these fields you have to define for each new resource where it can be equipped. 
Have I got that correctly?

steven2308: If you're setting either fixed or slot parts, there has to be a base. Otherwise it can be address 0.
The fixed parts, are the parts this resource will use. The slot parts, are the ones it can receive. (Both should be defined in the same base).
If you want to add a single resource with nothing from the base nor expected to equip something, you can set just the id and metadataURI, the rest would be empty.

stevyhacker: I see, that sounds good, thanks.

stevyhacker (approved):

PR #107, opened 2022-11-09
Removes index from multiresource accept/reject operations

steven2308:
This increases on average 25% the cost of adding resource to token and 7% the cost of accepting/rejecting.
In exchange, the contract size is reduced and the interface is simplified a lot for the user.
Updates MultiResourceRenderUtils and docs accordingly.


steven2308:
I'm attaching the change on gas for the affected operations. On `rejectAllResources` we're cleaning the index mapping since the for loop was already there. This can be avoided to save gas and would be harmless.
image

steven2308:
We won't be merging this do the increase in size and gas, but we might provide an extension which keeps tracks of indices using hooks, and implements methods with no indices, doing the call internally.

PR #110, opened 2022-11-10
Removes index on accept/unnest child

steven2308:
To achieve this, we need to keep track of children per index or both pending and active arrays.
To keep track, we need then 2 mappings of token Id -> Child address -> Child Id -> Index.
This change increases the contract size by ~0.201 kb, and increases the gas of `nestMint`, `acceptChild` and `unnestChild`.

Mappings are not being cleaned on the `rejectAllChildren` operation to avoid more gas and size. This is not desired but is harmless since we always check that the child on the found index is the intended one.

To avoid a bigger increase in size, we removed the childOf and pendingChildOf views, forcing users to either retrieve all children always, or to add custom methods to get the functionality if needed. To allow the latter, both `_activeChildren` and `_pendingChildren` arrays were made internal instead of private.

image


contracts/RMRK/nesting/RMRKNesting.sol
```diff
@@ -49,16 +49,24 @@ contract RMRKNesting is Context, IERC165, IERC721, IRMRKNesting, RMRKCore {
...
+    mapping(uint256 => Child[]) internal _pendingChildren;
 
     // Mapping of child token address to child token Id to whether they are pending or active on any token
     // We might have a first extra mapping from token Id, but since the same child cannot be
     // nested into multiple tokens we can strip it for size/gas savings.
     mapping(address => mapping(uint256 => uint256)) private _childIsInActive;
 
+    // Mapping of tokenId to childAddress to childId to index on the _pendingChildren array
+    mapping(uint256 => mapping(address => mapping(uint256 => uint256)))
+        private _pendingChildrenIndex;
```
stevyhacker, line 64: This doesn't look pretty but it works 😁

steven2308: Solidity in a sentence :laughing:

steven2308:
We won't be merging this do the increase in size and gas, but we might provide an extension which keeps tracks of indices using hooks, and implements methods with no indices, doing the call internally.

PR #114, opened 2022-11-22
repurposes unnestChild into transferChild

steven2308:
1. Makes it possible to unnest from a token directly to another token.
2. Renames `unnestChild` to `transferChild` since it better represents its wider reach. With transfer child you can do several operations which could have different buttons/actions for the user to make it simpler for them.
- Reject Child: from pending, to address 0
- Abandon Child: from active, to address 0
- Unnest Child: from pending or active, to the root owner
- Transfer to Address: from pending or active to any address. Should be an EoA or an `ERC721Receiver`.
- Transfer to Token: from pending or active to a collection + `tokenId` combination. Should be a `IRMRKNestable`.
3. Adds custom data parameter on `nestTransfer`, `transferChild` and `addChild`. This a sort of trick that  `ERC721` has, which allows to send custom data to the receiver, adding


test/extensions/soulbound.ts
```diff
@@ -136,15 +136,15 @@ describe('RMRKSoulbound exempt', async function () {
     expect(await token.directOwnerOf(tokenId)).eql([token.address, bn(otherTokenId), true]);
   });
 
-  it('can unnest child if soulbound exempt', async function () {
+  it('can transfer child if soulbound exempt', async function () {
```
ThunderDeliverer, line 139:
```suggestion
  it('Can transfer child if soulbound exempt', async function () {
```

test/extensions/soulbound.ts
```diff
@@ -211,13 +211,13 @@ async function shouldBehaveLikeSoulboundNestable() {
     ).to.be.revertedWithCustomError(soulbound, 'RMRKCannotTransferSoulbound');
   });
 
-  it('cannot unnest', async function () {
+  it('cannot transfer child', async function () {
```
ThunderDeliverer, line 214:
```suggestion
  it('Cannot transfer child', async function () {
```

ThunderDeliverer (commented):
Managed to finish this now. We can apply the changes in the new branch 😄 

They mostly address NatSpec format and some QoL improvements for tests

steven2308:
All tests are actually start with lowercase, I read them like "it cannot transfer child", so uppercase would be weird.

ThunderDeliverer:
> All tests are actually start with lowercase, I read them like "it cannot transfer child", so uppercase would be weird.

Makes sense 👌 There are some capitalized examples in tests though, we might want to downcap them then 🤔 

I can update the NatSpec if you'd like 🤔

steven2308:
I applied all the fixes, and also made sure capitalization is consistent on tests.
Thanks!

PR #118, opened 2022-11-24
Merges fixed and slot parts on Equippable

steven2308:
The complexity of telling them apart is delegated to EquipRenderUtils.

### DIFF
--- a/contracts/RMRK/nestable/RMRKNestable.sol
+++ b/contracts/RMRK/nestable/RMRKNestable.sol
@@ -930,9 +930,6 @@ contract RMRKNestable is Context, IERC165, IERC721, IRMRKNestable, RMRKCore {
         address childAddress,
         uint256 childId
     ) internal virtual {
-        if (pendingChildrenOf(parentId).length <= childIndex)
-            revert RMRKPendingChildIndexOutOfRange();
-
         Child memory child = pendingChildOf(parentId, childIndex);
         _checkExpectedChild(child, childAddress, childId);
         if (_childIsInActive[childAddress][childId] != 0)
--- a/contracts/RMRK/utils/RMRKEquipRenderUtils.sol
+++ b/contracts/RMRK/utils/RMRKEquipRenderUtils.sol
@@ -32,8 +32,7 @@ contract RMRKEquipRenderUtils {
         uint16 priority;
         address baseAddress;
         string metadata;
-        uint64[] fixedParts;
-        uint64[] slotParts;
+        uint64[] partIds;
     }
 
     /**
@@ -54,8 +53,7 @@ contract RMRKEquipRenderUtils {
         uint64 replacesAssetWithId;
         address baseAddress;
         string metadata;
-        uint64[] fixedParts;
-        uint64[] slotParts;
+        uint64[] partIds;
     }
 
     /**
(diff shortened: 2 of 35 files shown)


contracts/RMRK/utils/RMRKEquipRenderUtils.sol
```diff
@@ -435,4 +423,50 @@ contract RMRKEquipRenderUtils {
...
+                ++i;
+            }
+        }
+
+        slotPartIds = new uint64[](numSlotParts);
+        fixedPartIds = new uint64[](numFixedParts);
+        uint256 slotPartsIndex;
+        uint256 fixedPartsIndex;
+
+        // This for loop is to actually fill the split arrays
```
ThunderDeliverer, line 458: I wish this could have been done in the previous loop, but I get that this is impossible 😬

contracts/RMRK/utils/RMRKEquipRenderUtils.sol
```diff
@@ -435,4 +423,50 @@ contract RMRKEquipRenderUtils {
             }
         }
     }
+
+    function splitSlotAndFixedParts(
```
ThunderDeliverer, line 427: Love it! 🥳

test/setup/equippableSlots.ts
```diff
@@ -150,14 +150,11 @@ async function setupContextForSlots(
...
-      base.address,
-      'ipfs:soldier/',
-      [partIdForBody], // Fixed parts
-      [partIdForWeapon, partIdForBackground], // Can receive these
-    );
+    await soldierEquip.addAssetEntry(soldierResId, 0, base.address, 'ipfs:soldier/', [
+      partIdForBody,
+      partIdForWeapon,
+      partIdForBackground,
+    ]);
```
ThunderDeliverer, line 157: Does this improve readability? It's ok, just wondering 🤔

contracts/RMRK/nestable/RMRKNestable.sol
```diff
@@ -930,9 +930,6 @@ contract RMRKNestable is Context, IERC165, IERC721, IRMRKNestable, RMRKCore {
         address childAddress,
         uint256 childId
     ) internal virtual {
-        if (pendingChildrenOf(parentId).length <= childIndex)
-            revert RMRKPendingChildIndexOutOfRange();
-
```
ThunderDeliverer, line 935: Why is this check removed?

ThunderDeliverer (commented):
I love this improvement. I'm just wondering about removing the check in the Nestable 🤔

contracts/RMRK/nestable/RMRKNestable.sol
steven2308: The next line calls `pendingChildOf` which does the same check again

test/setup/equippableSlots.ts
steven2308: It's more evident if they are apart but it's a fair compromise. In any new sample or implementation I will simply include either "Part" or "Slot" on the variable name for the part.

contracts/RMRK/nestable/RMRKNestable.sol
ThunderDeliverer: Ok, then this is definitely a good catch 🔥

ThunderDeliverer (approved):
🚀

PR #120, opened 2022-11-28
Data on nesting hooks

steven2308:
I think it's useful to include data on hooks so people can actually use it for their custom purposes without having to rewrite the methods. It increases only 0.055 to the size on the biggest contract.


ThunderDeliverer (approved):
I agree, since the change is not that big in file size increase, but could potentially increase the quality of user experience, we should support it 👌

stevyhacker (approved):
Also agree with your point, seems like a useful addition without costing us too much.

PR #123, opened 2022-12-02
Add type safety to contract fields

stevyhacker:
Updated configs for ts/eslint
Added type safety to contracts where appropriate


steven2308 (approved):
Awesome!
I just noticed that some quotes were changed from single to double, while others remain single. 
I'll see what prettier does after merge.

PR #124, opened 2022-12-04
Verify and update NatSpec comments

ThunderDeliverer:
The NatSpec comments of the contracts were verified to include all of the parameters and appended where there were missing.

Some of the comments diverged from the implementation, so this was fixed as well.

The documentation was generated after the changes.


contracts/RMRK/nestable/IRMRKNestable.sol
```diff
@@ -185,7 +191,8 @@ interface IRMRKNestable is IERC165 {
     /**
      * @notice Used to reject all pending children of a given parent token.
      * @dev Removes the children from the pending array mapping.
-     * @dev The children's ownership structures are not updated.
+     * @dev This does not update the ownership storage data on children. If necessary, ownership can be reclaimed by the
+     *  rootOwner of the previous parent.
```
steven2308, line 195: I'm not sure this is worth adding here.
The thing is, it does not updated it because from the child's perspective this guy is already the parent.

docs/RMRK/access/OwnableLock.md
```diff
@@ -63,7 +63,7 @@ Used to check if the address is one of the contributors.
 
 | Name | Type | Description |
 |---|---|---|
-| _0 | bool | Boolean value indicating whether the address is a contributor or not |
+| _0 | bool | bool Boolean value indicating whether the address is a contributor or not |
```
steven2308, line 66: This "bool Boolean" on the description is weird, but I don't see the file changing :thinking:

contracts/implementations/RMRKNestableImpl.sol
```diff
@@ -12,6 +12,18 @@ error RMRKMintUnderpriced();
...
+     * @notice Used to initialize the smart contract.
+     * @param name_ Name of the token collection
+     * @param symbol_ Symbol of the token collection
+     * @param maxSupply_ Maximum supply of tokens in the collection
+     * @param pricePerMint_ Minting price of a token represented in the smallest denomination of the native currency
+     * @param collectionMetadata_ The collection metadata URI
+     * @param tokenURI_ The base URI of the token metadata
+     * @param royaltyRecipient The recipient of resale royalties
+     * @param royaltyPercentageBps The percentage of resale value to be allocated to the `royaltyRecipient` expressed in
+     *  the basis points
```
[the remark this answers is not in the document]
ThunderDeliverer: Yeah, I think that `in basis points` does the trick 👌

contracts/RMRK/nestable/IRMRKNestable.sol
ThunderDeliverer: I'd leave it. It might be a bit redundant, but might help someone when reading through the docs 🤔

PR #125, opened 2022-12-05
Add token properties extension

stevyhacker:
Simple on-chain tokenId-based properties extension with support for all data types indexed by numbers.

### DIFF
--- a/contracts/RMRK/extension/RMRKTokenProperties.sol
+++ b/contracts/RMRK/extension/RMRKTokenProperties.sol
@@ -9,4 +9,110 @@ pragma solidity ^0.8.16;
  */
 contract RMRKTokenProperties {
 
+    mapping(uint256 => mapping(uint256 => string)) private _stringProperties;
+    mapping(uint256 => mapping(uint256 => uint256)) private _uintProperties;
+    mapping(uint256 => mapping(uint256 => bool)) private _boolProperties;
+    mapping(uint256 => mapping(uint256 => address)) private _addressProperties;
+    mapping(uint256 => mapping(uint256 => bytes)) private _bytesProperties;
+
+    /**
+     * @notice Used to retrieve the token properties.
+     * @param tokenId The token ID
+     * @param index The index of the property
+     * @return string The value of the string property
+     */
+    function getStringTokenProperty(uint256 tokenId, uint256 index) external view returns (string memory) {
+        return _stringProperties[tokenId][index];
+    }
+
+    /**
+     * @notice Used to retrieve the token properties.
+     * @param tokenId The token ID
+     * @param index The index of the property
+     * @return uint256 The value of the uint property
+     */
+    function getUintTokenProperty(uint256 tokenId, uint256 index) external view returns (uint256) {
+        return _uintProperties[tokenId][index];
+    }
+
+    /**
+     * @notice Used to retrieve the token properties.
+     * @param tokenId The token ID
+     * @param index The index of the property
+     * @return bool The value of the bool property
+     */
+    function getBoolTokenProperty(uint256 tokenId, uint256 index) external view returns (bool) {
+        return _boolProperties[tokenId][index];
+    }
+
+    /**
+     * @notice Used to retrieve the token properties.
+     * @param tokenId The token ID
+     * @param index The index of the property
+     * @return address The value of the address property
+     */
+    function getAddressTokenProperty(uint256 tokenId, uint256 index) external view returns (address) {
+        return _addressProperties[tokenId][index];
+    }
+
+    /**
+     * @notice Used to retrieve the token properties.
+     * @param tokenId The token ID
+     * @param index The index of the property
+     * @return bytes The value of the bytes property
+     */
+    function getBytesTokenProperty(uint256 tokenId, uint256 index) external view returns (bytes memory) {
+        return _bytesProperties[tokenId][index];
+    }
+
+    /**
+     * @notice Used to set a number property.
+     * @param tokenId The token ID
+     * @param index The property index
+     * @param value The property value
+     */
+    function setIntProperty(uint256 tokenId, uint256 index, uint256 value) external {
+        _uintProperties[tokenId][index] = value;
+    }
+
+    /**
+    * @notice Used to set a string property.
+    * @param tokenId The token ID
+    * @param index The property index
+    * @param value The property value
+    */
+    function setStringProperty(uint256 tokenId, uint256 index, string memory value) external {
+        _stringProperties[tokenId][index] = value;
+    }
+
+    /**
+     * @notice Used to set a boolean property.
+     * @param tokenId The token ID
+     * @param index The property index
+     * @param value The property value
+     */
+    function setBoolProperty(uint256 tokenId, uint256 index, bool value) external {
+        _boolProperties[tokenId][index] = value;
+    }
+
+    /**
+     * @notice Used to set an bytes property.
+     * @param tokenId The token ID
+     * @param index The property index
+     * @param value The property value
+     */
+    function setBytesProperty(uint256 tokenId, uint256 index, bytes memory value) external {
+        _bytesProperties[tokenId][index] = value;
+    }
+
+    /**
+    * @notice Used to set an address property.
+     * @param tokenId The token ID
+     * @param index The property index
+     * @param value The property value
+     */
+    function setAddressProperty(uint256 tokenId, uint256 index, address value) external {
+        _addressProperties[tokenId][index] = value;
+    }
+
 }
--- a/docs/RMRK/extension/RMRKTokenProperties.md
+++ b/docs/RMRK/extension/RMRKTokenProperties.md
@@ -8,5 +8,213 @@ Smart contract of the RMRK Token properties module.
 
 
 
+## Methods
+
+### getAddressTokenProperty
+
+```solidity
+function getAddressTokenProperty(uint256 tokenId, uint256 index) external view returns (address)
+```
+
+Used to retrieve the token properties.
+
+
+
+#### Parameters
+
+| Name | Type | Description |
+|---|---|---|
+| tokenId | uint256 | The token ID |
+| index | uint256 | The index of the property |
+
+#### Returns
+
+| Name | Type | Description |
+|---|---|---|
+| _0 | address | address The value of the address property |
+
+### getBoolTokenProperty
+
+```solidity
+function getBoolTokenProperty(uint256 tokenId, uint256 index) external view returns (bool)
+```
+
+Used to retrieve the token properties.
+
+
+
+#### Parameters
+
+| Name | Type | Description |
+|---|---|---|
+| tokenId | uint256 | The token ID |
+| index | uint256 | The index of the property |
+
+#### Returns
+
+| Name | Type | Description |
+|---|---|---|
+| _0 | bool | bool The value of the bool property |
+
+### getBytesTokenProperty
+
+```solidity
+function getBytesTokenProperty(uint256 tokenId, uint256 index) external view returns (bytes)
+```
+
+Used to retrieve the token properties.
+
+
+
+#### Parameters
+
+| Name | Type | Description |
+|---|---|---|
+| tokenId | uint256 | The token ID |
+| index | uint256 | The index of the property |
+
+#### Returns
+
+| Name | Type | Description |
+|---|---|---|
+| _0 | bytes | bytes The value of the bytes property |
+
+### getStringTokenProperty
+
+```solidity
+function getStringTokenProperty(uint256 tokenId, uint256 index) external view returns (string)
+```
+
+Used to retrieve the token properties.
+
+
+
+#### Parameters
+
+| Name | Type | Description |
+|---|---|---|
+| tokenId | uint256 | The token ID |
+| index | uint256 | The index of the property |
+
+#### Returns
+
+| Name | Type | Description |
+|---|---|---|
+| _0 | string | string The value of the string property |
+
+### getUintTokenProperty
+
+```solidity
+function getUintTokenProperty(uint256 tokenId, uint256 index) external view returns (uint256)
+```
+
+Used to retrieve the token properties.
+
+
+
+#### Parameters
+
+| Name | Type | Description |
+|---|---|---|
+| tokenId | uint256 | The token ID |
+| index | uint256 | The index of the property |
+
+#### Returns
+
+| Name | Type | Description |
+|---|---|---|
+| _0 | uint256 | uint256 The value of the uint property |
+
+### setAddressProperty
+
+```solidity
+function setAddressProperty(uint256 tokenId, uint256 index, address value) external nonpayable
+```
+
+Used to set an address property.
+
+
+
+#### Parameters
+
+| Name | Type | Description |
+|---|---|---|
+| tokenId | uint256 | The token ID |
+| index | uint256 | The property index |
+| value | address | The property value |
+
+### setBoolProperty
+
+```solidity
+function setBoolProperty(uint256 tokenId, uint256 index, bool value) external nonpayable
+```
+
+Used to set a boolean property.
+
+
+
+#### Parameters
+
+| Name | Type | Description |
+|---|---|---|
+| tokenId | uint256 | The token ID |
+| index | uint256 | The property index |
+| value | bool | The property value |
+
+### setBytesProperty
+
+```solidity
+function setBytesProperty(uint256 tokenId, uint256 index, bytes value) external nonpayable
+```
+
+Used to set an bytes property.
+
+
+
+#### Parameters
+
+| Name | Type | Description |
+|---|---|---|
+| tokenId | uint256 | The token ID |
+| index | uint256 | The property index |
+| value | bytes | The property value |
+
+### setIntProperty
+
+```solidity
+function setIntProperty(uint256 tokenId, uint256 index, uint256 value) external nonpayable
+```
+
+Used to set a number property.
+
+
+
+#### Parameters
+
+| Name | Type | Description |
+|---|---|---|
+| tokenId | uint256 | The token ID |
+| index | uint256 | The property index |
+| value | uint256 | The property value |
+
+### setStringProperty
+
+```solidity
+function setStringProperty(uint256 tokenId, uint256 index, string value) external nonpayable
+```
+
+Used to set a string property.
+
+
+
+#### Parameters
+
+| Name | Type | Description |
+|---|---|---|
+| tokenId | uint256 | The token ID |
+| index | uint256 | The property index |
+| value | string | The property value |
+
+
 
 


steven2308 (changes requested):
It's hard to see how useful this is without tests.
I kind of expected properties to be accessed by name instead of an index, but I don't think an string can be a key on a mapping, or can it? What about bytes? 
If key must be an int I would call it Id instead of index, since it's a mapping and not an array, and it makes more sense also.

I'd rename `setIntProperty` -> `setUintProperty`

contracts/RMRK/extension/RMRKTokenProperties.sol
```diff
@@ -9,4 +9,110 @@ pragma solidity ^0.8.16;
...
 contract RMRKTokenProperties {
 
+    mapping(uint256 => mapping(uint256 => string)) private _stringProperties;
+    mapping(uint256 => mapping(uint256 => uint256)) private _uintProperties;
+    mapping(uint256 => mapping(uint256 => bool)) private _boolProperties;
+    mapping(uint256 => mapping(uint256 => address)) private _addressProperties;
+    mapping(uint256 => mapping(uint256 => bytes)) private _bytesProperties;
+
+    /**
+     * @notice Used to retrieve the token properties.
```
ThunderDeliverer, line 19:
```suggestion
     * @notice Used to retrieve the string type token properties.
```

contracts/RMRK/extension/RMRKTokenProperties.sol
```diff
@@ -9,4 +9,110 @@ pragma solidity ^0.8.16;
...
+     * @param tokenId The token ID
+     * @param index The index of the property
+     * @return string The value of the string property
+     */
+    function getStringTokenProperty(uint256 tokenId, uint256 index) external view returns (string memory) {
+        return _stringProperties[tokenId][index];
+    }
+
+    /**
+     * @notice Used to retrieve the token properties.
```
ThunderDeliverer, line 29:
```suggestion
     * @notice Used to retrieve the uint type token properties.
```

contracts/RMRK/extension/RMRKTokenProperties.sol
```diff
@@ -9,4 +9,110 @@ pragma solidity ^0.8.16;
...
+     * @param tokenId The token ID
+     * @param index The index of the property
+     * @return uint256 The value of the uint property
+     */
+    function getUintTokenProperty(uint256 tokenId, uint256 index) external view returns (uint256) {
+        return _uintProperties[tokenId][index];
+    }
+
+    /**
+     * @notice Used to retrieve the token properties.
```
ThunderDeliverer, line 39:
```suggestion
     * @notice Used to retrieve the bool type token properties.
```

contracts/RMRK/extension/RMRKTokenProperties.sol
```diff
@@ -9,4 +9,110 @@ pragma solidity ^0.8.16;
...
+     * @param tokenId The token ID
+     * @param index The index of the property
+     * @return bool The value of the bool property
+     */
+    function getBoolTokenProperty(uint256 tokenId, uint256 index) external view returns (bool) {
+        return _boolProperties[tokenId][index];
+    }
+
+    /**
+     * @notice Used to retrieve the token properties.
```
ThunderDeliverer, line 49:
```suggestion
     * @notice Used to retrieve the address type token properties.
```

contracts/RMRK/extension/RMRKTokenProperties.sol
```diff
@@ -9,4 +9,110 @@ pragma solidity ^0.8.16;
...
+     * @param tokenId The token ID
+     * @param index The index of the property
+     * @return address The value of the address property
+     */
+    function getAddressTokenProperty(uint256 tokenId, uint256 index) external view returns (address) {
+        return _addressProperties[tokenId][index];
+    }
+
+    /**
+     * @notice Used to retrieve the token properties.
```
ThunderDeliverer, line 59:
```suggestion
     * @notice Used to retrieve the bytes type token properties.
```

contracts/RMRK/extension/RMRKTokenProperties.sol
```diff
@@ -9,4 +9,110 @@ pragma solidity ^0.8.16;
...
+     * @param tokenId The token ID
+     * @param index The property index
+     * @param value The property value
+     */
+    function setBytesProperty(uint256 tokenId, uint256 index, bytes memory value) external {
+        _bytesProperties[tokenId][index] = value;
+    }
+
+    /**
+    * @notice Used to set an address property.
```
ThunderDeliverer, line 109:
```suggestion
     * @notice Used to set an address property.
```

ThunderDeliverer (changes requested):
Have you considered having an array of properties and then only assigning the IDs of these resources? So instead of:

```
mapping(uint256 => mapping(uint256 => string)) private _stringProperties;
```

This would be split into two parts:

```
mapping(uint256 => string) private stringProperties;
mapping(uint256 => mapping(uint256 => stringProperties)) private _stringProperties;
```

I think this could be beneficial as the string would have to be stored only once. This would be much cheaper for the maintainer.

As proprieties usually repeat themselves in a collection it might make sense.

I'd suggest using this approach for strings, addresses and bytes (it doesn't make sense to use it for booleans and integers) 🤔

stevyhacker:
> Have you considered having an array of properties and then only assigning the IDs of these resources? So instead of:
> 
> ```
> mapping(uint256 => mapping(uint256 => string)) private _stringProperties;
> ```
> 
> This would be split into two parts:
> 
> ```
> mapping(uint256 => string) private stringProperties;
> mapping(uint256 => mapping(uint256 => stringProperties)) private _stringProperties;
> ```

I agree it would be beneficial to save on storage costs when applying a token property on multiple tokens this way. Only issue is that this syntax is not valid
mapping(uint256 => string) private stringProperties;
mapping(uint256 => mapping(uint256 => stringProperties)) private _stringProperties;
It would go something like this and it looks a bit ugly:
    mapping(uint256 => mapping(string => mapping(uint256 => string))) private _stringProperties;

stevyhacker:
I've changed uint256 indexes to use a string key-value store and will add tests next.

ThunderDeliverer:
> > Have you considered having an array of properties and then only assigning the IDs of these resources? So instead of:
> > ```
> > mapping(uint256 => mapping(uint256 => string)) private _stringProperties;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This would be split into two parts:
> > ```
> > mapping(uint256 => string) private stringProperties;
> > mapping(uint256 => mapping(uint256 => stringProperties)) private _stringProperties;
> > ```
> 
> I agree it would be beneficial to save on storage costs when applying a token property on multiple tokens this way. Only issue is that this syntax is not valid
> 
> ```
> mapping(uint256 => string) private stringProperties;
> mapping(uint256 => mapping(uint256 => stringProperties)) private _stringProperties;
> ```
> 
> It would go something like this (with string keys) and it looks a bit ugly: ` mapping(uint256 => mapping(string => mapping(uint256 => string))) private _stringProperties;`

How about a single mapping-value struct?

So something akin to:

```solidity

    struct StringProperties {
        mapping(uint256 => StringProperties) stringProperties;
    }

    mapping(uint256 => mapping(string => StringProperties)) private _stringProperties;
```

I can't decide whether this is an improvement or not. It adds a bit more complexity, but also adds the gas saving feature we are discussing 🤔

contracts/RMRK/extension/tokenProperties/RMRKTokenProperties.sol
```diff
@@ -0,0 +1,144 @@
...
+pragma solidity ^0.8.16;
+
+import "./IRMRKTokenProperties.sol";
+
+/**
+ * @title RMRKTokenProperties
+ * @author RMRK team
+ * @notice Smart contract of the RMRK Token properties module.
+ */
+contract RMRKTokenProperties is IRMRKTokenProperties {
```
steven2308, line 12: I'm a bit confused by this class, I think it should be a mock since it has no gating.
The strategy we use for extensions is to have:
1. An interface
2. A core implementation, non opinionated. Here all the variables are defined as private with public getters and only internal setters. This is the one a custom implementer would extend.
3. A mock, with gated or not external setters which call the internal ones, this is the one we use for tests.
4. In some cases, an implementation, which adds gates. (I wouldn't add it for this one)

I feel this class is mixing 2 and 3. I'm open to discussion.

steven2308 (commented):
I'm not sure I understand why string, address and bytes properties are stored using a double mapping. Is it not possible to do something like this?
```
    mapping(uint256 => mapping(string => address)) private _addressProperties
```

contracts/RMRK/extension/tokenProperties/RMRKTokenProperties.sol
```diff
@@ -0,0 +1,144 @@
...
+
+    uint256 private totalStringProperties;
+    mapping(uint256 => string) stringProperties;
+    mapping(uint256 => mapping(string => uint256)) private _stringProperties;
+
+    uint256 private totalAddressProperties;
+    mapping(uint256 => address) addressProperties;
+    mapping(uint256 => mapping(string => uint256)) private _addressProperties;
```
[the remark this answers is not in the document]
steven2308: I just saw this was Jan's  proposal, call to defend it @ThunderDeliverer :sweat_smile:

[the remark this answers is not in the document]
ThunderDeliverer: String is easy, it's variable length and can consume a looot of resources in order to fully set it. You'd probably be burning through gas like your life depends on it.

Having this approach to strings allows for reuse of the same value on multiple occasions (which is quire common for large generative collections).

Address and fixed size bytes don't benefit as much (and one might argue that using an additional mapping might be more expensive), but I still feel for the continuity sake this is a good approach.

I'm open on rolling this back for addresses and bytes, but am confident this is the best approach for the strings.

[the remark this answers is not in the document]
steven2308: Ok I see the point now of mapping string keys to indices/ids. So you should only add a new index/id if that key does not exist already, right? 
I think we're missing that check.

Now I don't understand why you say you're convinced this is the best approach only for string values, if this affects specifically the key which is a string in every case. So shouldn't we actually do that for all properties, including uint and bool?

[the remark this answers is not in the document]
ThunderDeliverer: Well, for uint, you are using an uint to fetch an uint (possibly of the same value), so might be more expensive than just having a direct mapping. For boolean values it's kind of the same thing, you should only need 2 or 3 (in case you consider ID of 0 as unset) values, which seems like more storage usage than just assigning it directly 🤔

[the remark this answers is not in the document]
ThunderDeliverer: This is even better, yes! instead of using string keys in every mapping, we just use them to get the intended IDs this makes the interface easier to use and understand, but keeps the cost as low as possible.

I'd use the pattern you are proposing for all of the values other than string. For those I'd still use another mapping, so one that uses indices to retrieve the string property values:
`mapping(uint256 => string) stringProperties;`
I still feel like this might be a great improvement:
`mapping(string => id) addressKeysToIds;`
And then the mapping would be:
`mapping(uint256 => mapping(uint256 => uint256)) private _stringValues;`

So to retrieve a string value:

```
id = addressKeysToIds;
index = _stringValues[tokenId][id];
return stringProperties[index];
```

This should be easy to justify and noone should have a problem with accepting a slightly different approach.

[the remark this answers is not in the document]
steven2308: I don't see how this is better for strings.
Also, let's say you have some key K assigned to id A (We agree K->A will be a thing for ever now).
If you 2 tokens use this key, I think you'd be forcing them to always have the same value. Unless what I'm assuming we agree is false.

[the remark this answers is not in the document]
ThunderDeliverer: Much like you could modify the value of `_addressValues`, you could do the same for `_stringValue`, just at a different point.

You change the `_addressValues` like this:
`_addressValues[tokenId][id] = newAddressValue;`

And you would change the string value like this:
`stringProperties[index] = newStringValue;`

It's just a different place where you can modify the value, but it is still modifiable if needed, or if you want to use another, already existing, string value, you just do:
`_stringValues[tokenId][id] = newIndex;`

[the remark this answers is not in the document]
steven2308: Then you would need to create a new id for the same key. Which seems bad IMO. And I still don't see what you gain for strings.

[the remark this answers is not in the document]
ThunderDeliverer: ID could remain the same only the string value accessible by the index or the index itself would be changed if you wish to do this. I don't see the need to change IDs 🤔

contracts/RMRK/extension/tokenProperties/RMRKTokenProperties.sol
stevyhacker: Ok so in this case you would prefer to call this RMRKTokenPropertiesMock ?

steven2308: Yeap that would be the easiest.

contracts/mocks/extensions/tokenProperties/RMRKTokenPropertiesMock.sol
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * @notice Used to set a number property.
+     * @param tokenId The token ID
+     * @param key The property key
+     * @param value The property value
+     */
+    function setUintProperty(uint256 tokenId, string memory key, uint256 value) external {
+        _keysToIds[key] = _totalProperties;
```
steven2308, line 82: You need to check if that key was already mapped to an Id for 2 reasons:
1. Optimization, same keys should use the same ID.
2. If it existed before you're rewriting it, and the previous value associated to that key won't be found because when reading you will point to a different id. I suggest using something like this on every setter to get the right Id:
```
function _getIdForKey(string memory key) internal returns (uint256) {
        if (_keysToIds[key] == 0) {
            _totalProperties += 1; // Add first so it never uses 0 as ID
            _keysToIds[key] = _totalProperties;
            return _totalProperties;
        }
        else {
            return _keysToIds[key];
        }
    } 
```

On setters you simply get the Id by `_getIdForKey(key)` without worrying if it existed or not.

stevyhacker: Added it now for all the setters except strings.

stevyhacker: As strings setter already checks for duplicate keys and values.

steven2308 (approved):
Looks good to me!

contracts/mocks/extensions/tokenProperties/RMRKTokenPropertiesMock.sol
```diff
@@ -2,29 +2,25 @@
...
-    mapping(uint256 => string) stringProperties;
-    mapping(uint256 => mapping(string => uint256)) private _stringProperties;
+    mapping(string => uint256) private _keysToIds;
+    uint256 private _totalProperties = 0;
 
-    uint256 private totalAddressProperties;
-    mapping(uint256 => address) addressProperties;
-    mapping(uint256 => mapping(string => uint256)) private _addressProperties;
+    mapping(string => uint256) private _stringValuesToIds;
+    mapping(uint256 => mapping(uint256 => string)) private _stringValues;
```
ThunderDeliverer, line 18: This seems different from what we agreed on the call 🤔 

I think this is in line with what we agreed on:

```suggestion
    mapping(uint256 => string) private dictionary;
    mapping(string => uint256) private reverseDictionary;
    mapping(uint256 => mapping(uint256 => uint256)) private _stringValues;
```

contracts/mocks/extensions/tokenProperties/RMRKTokenPropertiesMock.sol
```diff
@@ -33,7 +29,7 @@ contract RMRKTokenProperties is IRMRKTokenProperties {
      * @return string The value of the string property
      */
     function getStringTokenProperty(uint256 tokenId, string memory key) external view returns (string memory) {
-        return stringProperties[_stringProperties[tokenId][key]];
+        return _stringValues[tokenId][_keysToIds[key]];
```
ThunderDeliverer, line 32: So based on what we agreed on, this should be:

```suggestion
        return dictionary[_stringValues[tokenId][_keysToIds[key]]];
```

contracts/mocks/extensions/tokenProperties/RMRKTokenPropertiesMock.sol
```diff
@@ -93,9 +89,22 @@ contract RMRKTokenProperties is IRMRKTokenProperties {
     * @param value The property value
     */
     function setStringProperty(uint256 tokenId, string memory key, string memory value) external {
-        _stringProperties[tokenId][key] = totalStringProperties;
-        stringProperties[totalStringProperties] = value;
-        totalStringProperties++;
+
```
ThunderDeliverer, line 92: Again, based on what we agreed on, this needs:

```suggestion
if(reverseDictionary[value] == 0) {
    dictionary[dictionary.length + 1] = value;
    reverseDictionary[value] = dictionary.length;
}
```

I just assumed that value 0 is empty (like we do for asset & child IDs), so that this is smaller.

contracts/mocks/extensions/tokenProperties/RMRKTokenPropertiesMock.sol
```diff
@@ -93,9 +89,22 @@ contract RMRKTokenProperties is IRMRKTokenProperties {
...
+        if (_stringValuesToIds[value] == 0 && _keysToIds[key] == 0) {
+            _keysToIds[key] = _totalProperties;
+            _stringValues[tokenId][_keysToIds[key]] = value;
```
ThunderDeliverer, line 95:
```suggestion
            _stringValues[tokenId][_keysToIds[key]] = reverseDictionary[value];
```

contracts/mocks/extensions/tokenProperties/RMRKTokenPropertiesMock.sol
```diff
@@ -2,29 +2,25 @@
...
-    uint256 private totalStringProperties;
-    mapping(uint256 => string) stringProperties;
-    mapping(uint256 => mapping(string => uint256)) private _stringProperties;
+    mapping(string => uint256) private _keysToIds;
+    uint256 private _totalProperties = 0;
 
-    uint256 private totalAddressProperties;
-    mapping(uint256 => address) addressProperties;
-    mapping(uint256 => mapping(string => uint256)) private _addressProperties;
+    mapping(string => uint256) private _stringValuesToIds;
```
ThunderDeliverer, line 17: I just realized that this was meant to be what I named `reverseDictionary`, so I'll stop adding suggestions now, but this should be accompanied by the opposite mapping as well and used in the `_stringValues`.

ThunderDeliverer (changes requested):
Mid-review I realised that half of what we agreed on was implemented, so I just added a comment explaining what is missing.

ThunderDeliverer:
Also, the docs were removed in one of the latest pushes and need to be regenerated and readded.

PR #126, opened 2022-12-05
Fixes max supply to total supply when contract is locked

steven2308:
Adds missing coverage on minting utils and fixes 2 bugs.


contracts/RMRK/access/Ownable.sol
```diff
@@ -66,7 +66,7 @@ contract Ownable is Context {
      * @param newOwner Address of the new owner's account
      */
     function transferOwnership(address newOwner) public virtual onlyOwner {
-        if (owner() == address(0)) revert RMRKNewOwnerIsZeroAddress();
```
stevyhacker, line 69: Good catch, it took me a while to see what's the problem here 😂

stevyhacker (approved):

PR #128, opened 2022-12-08
Token URI can be enumerated on implementations

steven2308:


contracts/implementations/RMRKEquippableImpl.sol
```diff
@@ -11,36 +12,37 @@ error RMRKMintUnderpriced();
...
-        string memory symbol,
-        uint256 maxSupply,
-        uint256 pricePerMint,
+        string memory name_,
+        string memory symbol_,
         string memory collectionMetadata_,
         string memory tokenURI_,
-        address royaltyRecipient,
-        uint256 royaltyPercentageBps //in basis points
+        InitData memory data
```
stevyhacker, line 38: I see we are back to using structs for params 😄

steven2308: Yeah, now every implementation has to use it, not only erc20s

PR #130, opened 2022-12-08
Lazy minting is now optional on implementations.

steven2308:
It can be disabled so only owner is allowed to mint.

### DIFF
--- a/contracts/RMRK/utils/RMRKMintingUtils.sol
+++ b/contracts/RMRK/utils/RMRKMintingUtils.sol
@@ -15,16 +15,23 @@ contract RMRKMintingUtils is OwnableLock {
     uint256 internal _totalSupply;
     uint256 internal _maxSupply;
     uint256 internal immutable _pricePerMint;
+    uint256 internal immutable _lazyMintingEnabled;
 
     /**
      * @notice Initializes the smart contract with a given maximum supply and minting price.
      * @param maxSupply_ The maximum supply of tokens to initialize the smart contract with
      * @param pricePerMint_ The minting price to initialize the smart contract with, expressed in the smallest
      *  denomination of the native currency of the chain to which the smart contract is deployed to
+     * @param lazyMintingEnabled_ Whether the token minting is lazy or not. If not, only the owner can mint
      */
-    constructor(uint256 maxSupply_, uint256 pricePerMint_) {
+    constructor(
+        uint256 maxSupply_,
+        uint256 pricePerMint_,
+        bool lazyMintingEnabled_
+    ) {
         _maxSupply = maxSupply_;
         _pricePerMint = pricePerMint_;
+        _lazyMintingEnabled = lazyMintingEnabled_ ? 1 : 0;
     }
 
     /**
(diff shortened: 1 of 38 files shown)


contracts/implementations/IRMRKInitData.sol
```diff
@@ -13,7 +13,8 @@ interface IRMRKInitData {
      * @notice Used to provide initialization data without running into stack too deep errors.
      * @return erc20TokenAddress Address of the ERC20 token to be used when initializing a smart contract that supports
      *  ERC20 pay module
-     * @return tokenUriIsEnumerable Weather the token URI is enumerable or not
+     * @return tokenUriIsEnumerable Whether the token URI is enumerable or not
```
ThunderDeliverer, line 16:
```suggestion
     * @return tokenUriIsEnumerable Boolean value signifying whether the token URI is enumerable or not
```

contracts/implementations/IRMRKInitData.sol
```diff
@@ -13,7 +13,8 @@ interface IRMRKInitData {
...
+     * @return lazyMintingEnabled Whether the token minting is lazy or not. If not, only the owner can mint
```
ThunderDeliverer, line 17:
```suggestion
     * @return lazyMintingEnabled Boolean value signifying whether the token minting is lazy or not. If not, only the owner can mint tokens
```

contracts/RMRK/utils/RMRKMintingUtils.sol
```diff
@@ -95,4 +110,13 @@ contract RMRKMintingUtils is OwnableLock {
     function _checkSaleIsOpen() private view {
         if (_totalSupply >= _maxSupply) revert RMRKMintOverMax();
     }
+
+    /**
+     * @notice Used to verify that that only the owner can mint tokens if lazy minting is not enabled.
+     * @dev In case lazy minting is not enable and msg sender is not the contract owner, the execution is reverted.
+     */
+    function _checkLazyMinting() internal view {
+        if (_lazyMintingEnabled == 0 && _msgSender() != owner())
```
ThunderDeliverer, line 119: I probably went through this check 10 times just to be sure it doesn't bite us in the ass sometime in the future 😂

test/implementations/erc20pay.ts
```diff
@@ -64,10 +67,31 @@ async function equippableFixture(): Promise<Contract> {
...
+}> {
+  const erc20Factory = await ethers.getContractFactory('ERC20Mock');
+  const erc20 = <ERC20Mock>await erc20Factory.deploy();
+  await erc20.deployed();
+
+  const lazyMintingEnabled = false; // Shadows module lazyMintingEnabled
+  const token = <RMRKMultiAssetImplErc20Pay>(
+    await singleFixtureWithArgs('RMRKMultiAssetImplErc20Pay', [
+      'MultiAsset',
+      'MR',
```
ThunderDeliverer, line 86: This is probably a remainder from the MultiResource era, maybe change it to `MA`? 😅 

```suggestion
      'MA',
```

ThunderDeliverer (approved):
Implementation wise I think we are golden and tests seem to provide full coverage. Would you ming applying doc changes before merge?

steven2308:
We'll discard this on favor of independent versions for pre minting

PR #131, opened 2022-12-09
Adss implementations with pre minting strategy and other improvements

steven2308:
On pre minting implementations, only owner can mint and there's no cost.
Moves IRMRKInitData import to abstract on implementations.
Removes redundant modifier and improves coverage.

### DIFF
--- a/contracts/implementations/RMRKNestableImpl.sol
+++ b/contracts/implementations/RMRKNestableImpl.sol
@@ -53,7 +53,7 @@ contract RMRKNestableImpl is RMRKAbstractNestableImpl {
     function mint(
         address to,
         uint256 numToMint
-    ) public payable virtual notLocked saleIsOpen {
+    ) public payable virtual notLocked {
         (uint256 nextToken, uint256 totalSupplyOffset) = _preMint(numToMint);
 
         for (uint256 i = nextToken; i < totalSupplyOffset; ) {
@@ -76,7 +76,7 @@ contract RMRKNestableImpl is RMRKAbstractNestableImpl {
         address to,
         uint256 numToMint,
         uint256 destinationId
-    ) public payable virtual notLocked saleIsOpen {
+    ) public payable virtual notLocked {
         (uint256 nextToken, uint256 totalSupplyOffset) = _preMint(numToMint);
 
         for (uint256 i = nextToken; i < totalSupplyOffset; ) {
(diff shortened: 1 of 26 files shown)


test/implementations/premint.ts
```diff
@@ -0,0 +1,155 @@
...
+  singleFixtureWithArgs,
+} from '../utils';
+import { Contract } from 'ethers';
+import { loadFixture } from '@nomicfoundation/hardhat-network-helpers';
+import { ERC20Mock } from '../../typechain-types';
+
+async function multiAssetFixture(): Promise<Contract> {
+  return await singleFixtureWithArgs('RMRKMultiAssetImplPreMint', [
+    'MultiAsset',
+    'MR',
```
ThunderDeliverer, line 18: Like I mentioned before. It would be cool to rename this to `MA`, for someone taking a look at our codebase, who is unfamiliar with the MultiResource, this could be confusing 🤔 

```suggestion
    'MA',
```

ThunderDeliverer: I'd suggest doing this for all instances.

contracts/implementations/RMRKNestableImpl.sol
```diff
@@ -54,7 +53,7 @@ contract RMRKNestableImpl is IRMRKInitData, RMRKAbstractNestableImpl {
     function mint(
         address to,
         uint256 numToMint
-    ) public payable virtual notLocked saleIsOpen {
```
ThunderDeliverer, line 57: Why is this removed? To be less opinionated or just to preserve contract size?

ThunderDeliverer (approved):
Am I correct to assume that minting in `erc20pay` is not `payable` any more because ERC20 tokens are meant to be used to pay for minting?

If all of the questions that I had were done intentionally, I think this is good to be merged 👌

test/implementations/premint.ts
steven2308: Will do on dev, thanks!

contracts/implementations/RMRKNestableImpl.sol
steven2308: `saleIsOpen` was checking that  `totaSupply <= maxSupply`. I noticed that the modifier was never used and when tracking it it happens that on minting we check that the current supply + the total being minted doesn't go over the max supply.  So the check on `saleIsOpen` is non sense, it will never be reached.

steven2308:
> Am I correct to assume that minting in `erc20pay` is not `payable` any more because ERC20 tokens are meant to be used to pay for minting?
> 
> If all of the questions that I had were done intentionally, I think this is good to be merged ok_hand

Yep, there was one of the implementations where I had forgotten to remove the payable which is not necessary.

PR #132, opened 2022-12-10
change compare condition from less or equal 0 to equal 0

BoynChan:
I found this code when I view it. But I don't think it needed to check if length of array is less than 0


steven2308:
Yep, that makes sense. Thanks!

PR #133, opened 2022-12-13
Improvements on token properties

steven2308:
Adds tests for duplicate keys, keys shared among different types and values shared among different string properties.
Caches string values to reduce gas when same value is used several times.
Splits minimal abstract implementation and mock to make it easier to extend.

### DIFF
--- a/contracts/RMRK/extension/tokenProperties/RMRKTokenProperties.sol
+++ b/contracts/RMRK/extension/tokenProperties/RMRKTokenProperties.sol
@@ -1,0 +1,199 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pragma solidity ^0.8.16;
+
+import "./IRMRKTokenProperties.sol";
+
+/**
+ * @title RMRKTokenProperties
+ * @author RMRK team
+ * @notice Smart contract of the RMRK Token properties module.
+ */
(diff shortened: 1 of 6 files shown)


contracts/RMRK/extension/tokenProperties/RMRKTokenProperties.sol
```diff
@@ -0,0 +1,199 @@
...
+pragma solidity ^0.8.16;
+
+import "./IRMRKTokenProperties.sol";
+
+/**
+ * @title RMRKTokenProperties
+ * @author RMRK team
+ * @notice Smart contract of the RMRK Token properties module.
+ */
+abstract contract RMRKTokenProperties is IRMRKTokenProperties {
```
stevyhacker, line 12: I have really grown to like how abstract contracts are useful. Great improvement over the previous version.

stevyhacker (approved):

contracts/RMRK/extension/tokenProperties/RMRKTokenProperties.sol
steven2308: For our use case of creating a standard + minimal + full implementation, it's a game changer.
Thanks!

PR #137, opened 2022-12-14
Regenerate docs on compile

ThunderDeliverer:
Added dodoc config, so that the documentation is regenerated every time the contracts are compiled.

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -62,6 +62,9 @@ const config: HardhatUserConfig = {
       goerli: process.env.ETHERSCAN_API_KEY || '', // Goerli Etherscan API Key
     },
   },
+  dodoc: {
+    runOnCompile: true,
+  },
 };
 
 export default config;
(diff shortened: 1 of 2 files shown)


stevyhacker (approved):

steven2308 (commented):
I'm pretty sure it was already doing it. 
Do you know if it's possible to skip it when running tests? It's a bit of a pain since it always regenerates so I'm commenting the import on hardhat config all the time

ThunderDeliverer:
#125 didn't have the docs, so I added this, to make sure the docs are compiled. Better to be redundant, than to miss it 🤔 The alternative is to disable it and just rely that we remember to regenerate the docs manually using the checklist in the PR template.. This could help us speed up the tests, while still reminding us to regenerate docs.

Maybe we could also add a custom script:
````json
    "prepare": "dodoc && prettier --write ."
````

If you have any additional ideas on what to add to this script, so that all of the final actions should be taken, please let me know.

PR #143, opened 2023-01-04
Emotable Extension

steven2308:
# Description

Adds interface, abstract implementation and mock implementation for emotable NFTs extension.
Also includes adding equip/unequip to other Equippables interface and bump to version 0.22.1 (by mistake)

# Checklist

- [Ok] Verified code additions
- [Ok] Updated NatSpec comments (if applicable)
- [Ok] Regenerated docs
- [Ok] Ran prettier
- [Ok] Added tests to fully cover the changes


contracts/RMRK/extension/emotable/RMRKEmotable.sol
```diff
@@ -0,0 +1,79 @@
...
+        if (currentVal != state) {
+            _beforeEmote(tokenId, emoji, state);
+            if (state) {
+                _emotesPerToken[tokenId][emoji] += 1;
+            } else {
+                _emotesPerToken[tokenId][emoji] -= 1;
+            }
+            _emotesPerAddress[msg.sender][tokenId][emoji] = state ? 1 : 0;
+            emit Emoted(msg.sender, tokenId, emoji, state);
+        }
```
ThunderDeliverer, line 56: I think this one could use an `else` clause:

```suggestion
        if (currentVal != state) {
            _beforeEmote(tokenId, emoji, state);
            if (state) {
                _emotesPerToken[tokenId][emoji] += 1;
            } else {
                _emotesPerToken[tokenId][emoji] -= 1;
            }
            _emotesPerAddress[msg.sender][tokenId][emoji] = state ? 1 : 0;
            emit Emoted(msg.sender, tokenId, emoji, state);
        } else {
            revert RMRKAlreadyEmoted();
        }
```

contracts/RMRK/extension/emotable/RMRKEmotable.sol
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * @notice Hook that is called before emote is added or removed.
+     * @param tokenId ID of the token being emoted
+     * @param emoji Unicode identifier of the emoji
+     * @param state whether to turn emote or undo. True for emote, false for undo
```
ThunderDeliverer, line 63:
```suggestion
     * @param state Boolean value signifying whether to emote (`true`) or undo (`false`) emote
```

test/extensions/emotable.ts
```diff
@@ -0,0 +1,89 @@
...
+  it('does not support other interfaces', async function () {
+    expect(await token.supportsInterface(IOtherInterface)).to.equal(false);
+  });
+
+  describe('With minted tokens', async function () {
+    beforeEach(async function () {
+      await token.mint(owner.address, tokenId);
+    });
+
+    it('can emote and undo emote', async function () {
```
ThunderDeliverer, line 53: I think we could split this example into `can emote` and `can undo emote`, since these are two actions that need to be verified 🤔

ThunderDeliverer (changes requested):
I was being a bit nitpicky with NatSpec comments. Sorry about that 😂 😇

contracts/RMRK/extension/emotable/RMRKEmotable.sol
steven2308: It was an option, but I don't see the harm on trying to reapply a state so I went for no revert.
Additionally, in this proposal you would raise `RMRKAlreadyEmoted` when state is off and you're setting off again. So the error doesn't make sense, we'd need to errors (more code and contract size) or a more generic. In any case I don't see a value against just ignoring.

test/extensions/emotable.ts
steven2308: We could.. but meh. Go ahead if you want :sweat_smile:

contracts/RMRK/extension/emotable/RMRKEmotable.sol
ThunderDeliverer: I think no revert path is good if we are certain that the function will be called directly and that no other actions will be taken afterwards. If there is a relatively high possibility of other on chain actions following this one in the same transaction, I think it would be useful to revert in order to use as little gas as possible. I agree about the error having to be more generic.

That said, I'm ok with the current implementation if you both agree with it 🦾

test/extensions/emotable.ts
ThunderDeliverer: Done 😅

ThunderDeliverer (approved):
I have no further comments 🚀

contracts/RMRK/extension/emotable/RMRKEmotable.sol
steven2308: If it's going to be called before something else (totally possible) the caller shouldn't assume such behavior. What if we just add to the docstring that will do nothing if state is already the same?  
@stevyhacker  what are your thought's on this?

stevyhacker: I am okay with the current implementation, emote function works like a toggle button in this case, no real benefit for raising an error/reverting here as this is a very simple operation.

stevyhacker (approved):
Looks pretty good to me!

contracts/RMRK/extension/emotable/RMRKEmotable.sol
ThunderDeliverer: Let's add `@dev` tag noting this down like you suggested @steven2308 and finalize this discussion 👌

PR #148, opened 2023-01-13
Replace dodoc with docgen

ThunderDeliverer:
# Description

As `dodoc` proved to result in some unexpected results and seems to have been abandoned, it was replaced with `docgen` and the documentation has been regenerated.

A thing to consider is that `docgen` doesn't generate the documentation for inherited functions and errors, so the documentation is not as extensive. On the other hand, it does generate docs for internal functions (which is really useful for more advanced implementations).

I'll mark this as a draft and explore alternative options and configurations.

# Actions

- Replaced `dodoc` with `docgen`
- Regenerated documentation
- Updated documentation of `VERSION` in `Core`
- Updated `prepare` script to use `docgen`

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier
- [N/A] Added tests to fully cover the changes


ThunderDeliverer:
I've spent the weekend mulling over the pros and cons of both.. I think that `docgen`'s inability to render the full documentation of the contract (it only renders documentation for the functions that are contained within the file, but doesn't include the inherited functions/errors/..) is worse than a few `undefined` values.

I've tried finding a maintained documentation generator, but there seems to be none. Another thing to note is that the `dodoc` only seems to have issues with custom input parameters; namely custom structs. I'll see if I can identify the issue and rectify it within the package. If this works, we would have our own, arguably the best, documentation generator. If not, we will have to decide which pains us less and use that one.

steven2308:
I agree on undefined being less of a problem. If we manage to find and fix the issue we can PR the dodoc repo. What's it written in?

ThunderDeliverer:
Upon weighing pros and cons of both, we decided to keep `dodoc` as our documentation generator.

PR #156, opened 2023-01-27
Add get top asset to render utils

ThunderDeliverer:
# Description

Yuri mentioned that when retrieving the top asset it is common to require ID, priority and metadata URI of said asset. A method to retrieve all of them in the same call was not yet present in the `RMRKMultiAssetRenderUtils`, so I added it.

# Actions

Added a `getTopAsset` method to return all of these values with one call.

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier


steven2308 (approved):
Nice!

PR #159, opened 2023-01-27
Fix ipfs gateway in deploy example

ilionic:


steven2308:
I removed the script so this is no longer needed.

PR #167, opened 2023-02-07
Removes allowance check on ERC20Pay versions since transfer will fail…

steven2308:
# Description

The transfer call will fail anyway and we save 0.177 in the ERC20Pay implementations

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier


stevyhacker (approved):
Small change but very worth it!

ThunderDeliverer:
Really nice! 👌

PR #168, opened 2023-02-07
Includes toZero parameter on the ChildTransferred event

steven2308:
# Description

Includes `toZero` parameter on the `ChildTransferred` event.
This is needed for indexers to identify if a child was removed or rejected, since `NestTransfer` event only emits when address is not 0.

# Actions

- Adds `toZero` on `ChildTransferred` Event
- Updates nestable tests which where checking for the event and args
- Adds test cases for removing and rejecting a child.

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

### DIFF
--- a/contracts/RMRK/nestable/IRMRKNestable.sol
+++ b/contracts/RMRK/nestable/IRMRKNestable.sol
@@ -91,7 +91,7 @@ interface IRMRKNestable is IERC165 {
      * @param childId ID of the child token in the child token's collection smart contract
      * @param fromPending A boolean value signifying whether the token was in the pending child tokens array (`true`) or
      *  in the active child tokens array (`false`)
-     * @param toZero A boolean value signifying whether the token is being transferred to the address zero (`true`) or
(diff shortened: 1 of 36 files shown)


contracts/RMRK/nestable/IRMRKNestable.sol
```diff
@@ -91,13 +91,16 @@ interface IRMRKNestable is IERC165 {
      * @param childId ID of the child token in the child token's collection smart contract
      * @param fromPending A boolean value signifying whether the token was in the pending child tokens array (`true`) or
      *  in the active child tokens array (`false`)
+     * @param toZero A boolean value signifying whether the token is being transferred to the address zero (`true`) or
```
ThunderDeliverer, line 94: I'd reword this just a little bit:

```suggestion
     * @param toZero A boolean value signifying whether the token is being transferred to the `0x0` address (`true`) or
```

ThunderDeliverer (approved):
Neat! 👌

PR #170, opened 2023-02-08
Implements BulkWriterPerCollection contract.

steven2308:
# Description

This contract is meant to exist 1 per NFT collection, the idea is to limit danger if the contract is somehow compromised. That however makes UX worse, since we need to deploy one per collection and users need to approve it too.

It's up for discussion

# Actions

- Implements replace equip
- Implements bulk equip

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

### DIFF
--- a/contracts/RMRK/utils/RMRKBulkWriterPerCollection.sol
+++ b/contracts/RMRK/utils/RMRKBulkWriterPerCollection.sol
@@ -3,6 +3,10 @@
 pragma solidity ^0.8.18;
 
 import "../equippable/IRMRKEquippable.sol";
+import "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+
+error RMRKCanOnlyDoBulkOperationsOnOwnedTokens();
+error RMRKCanOnlyDoBulkOperationsWithOneTokenAtATime();
 
 /**
  * @title RMRKBulkWriterPerCollection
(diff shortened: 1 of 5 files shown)


contracts/RMRK/utils/RMRKBulkWriterPerCollection.sol
```diff
@@ -0,0 +1,134 @@
...
+            IRMRKEquippable(_collection).unequip(
+                tokenId,
+                unequips[i].assetId,
+                unequips[i].slotPartId
+            );
+        }
+        length = equips.length;
+        for (uint256 i = 0; i < length; i++) {
+            if (equips[i].tokenId != tokenId) {
+                revert RMRKCanOnlyDoBulkOperationsWithOneTokenAtATime();
```
stevyhacker, line 117: I think from a user standpoint, many will request/expect the ability to do bulk operations with multiple tokens.

steven2308: I initially did it that way, but that implied an external call on every step of the loop which felt really dumb.
I expect this to be called from a token detail page where you would do several operations on the same token. I don't think it would be practical for a UI to track changes from different tokens, do you?

stevyhacker: We can always make a 2nd version of this if need be so I guess it's fine for now.

stevyhacker (approved):

PR #171, opened 2023-02-15
Allows to set multiple tokenIds in AssetAddedToToken event

steven2308:
Warning: It increases the contract size by 0.129.

# Description

Changes the event parameter from `uint256 indexed tokenId` to `uint256 tokenIds[]`

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

### DIFF
--- a/contracts/RMRK/multiasset/IRMRKMultiAsset.sol
+++ b/contracts/RMRK/multiasset/IRMRKMultiAsset.sol
@@ -19,12 +19,12 @@ interface IRMRKMultiAsset is IERC165 {
     /**
      * @notice Used to notify listeners that an asset object at `assetId` is added to token's pending asset
      *  array.
-     * @param tokenId ID of the token that received a new pending asset
+     * @param tokenIds An array of token IDs that received a new pending asset
      * @param assetId ID of the asset that has been added to the token's pending assets array
      * @param replacesId ID of the asset that would be replaced
      */
     event AssetAddedToToken(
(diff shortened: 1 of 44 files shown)


contracts/RMRK/multiasset/IRMRKMultiAsset.sol
```diff
@@ -19,12 +19,12 @@ interface IRMRKMultiAsset is IERC165 {
     /**
      * @notice Used to notify listeners that an asset object at `assetId` is added to token's pending asset
      *  array.
-     * @param tokenId ID of the token that received a new pending asset
+     * @param tokenIds An array of token IDs that received a new pending asset
      * @param assetId ID of the asset that has been added to the token's pending assets array
      * @param replacesId ID of the asset that would be replaced
      */
     event AssetAddedToToken(
```
ThunderDeliverer, line 26: I'd also rename the event to `AssetAddedToTokens` in order to reflect that this can now impact multiple tokens at once 🤔

stevyhacker: Agree, it should be plural because as it is now is a bit misleading.

steven2308: I agree with your agreement. :smile: 
Change is done.

ThunderDeliverer (approved):
🎆

PR #172, opened 2023-02-15
Uses memory parameters instead of calldata to receive partIds

steven2308:
# Description

Any use case which builds the list of parts with code will be blocked otherwise, since this happens in the core implementation and all variables are private. The only workaround would be to modify the installed package which is a really bad practice.

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier


stevyhacker (approved):

ThunderDeliverer (approved):
🚀

PR #174, opened 2023-02-22
Emits events before the hooks.

steven2308:
In few cases, the events where being triggered after the hook which is not a great practice and was not consistent with the rest of the code

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier


ThunderDeliverer (approved):
⭐

PR #175, opened 2023-03-01
Removes non reentrant guard from equipping.

steven2308:
# Description

Ideally we would keep the reentrancy guard on equip since there's an external call before the state is updated.  However, impact is so low and contract size so tight that I propose to remove it. It saves 0.056 on contract size.

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

### DIFF
--- a/contracts/RMRK/equippable/RMRKEquippable.sol
+++ b/contracts/RMRK/equippable/RMRKEquippable.sol
@@ -8,7 +8,6 @@ import "../catalog/IRMRKCatalog.sol";
 import "../library/RMRKLib.sol";
 import "../multiasset/AbstractMultiAsset.sol";
 import "../nestable/RMRKNestable.sol";
-import "../security/ReentrancyGuard.sol";
 import "./IRMRKEquippable.sol";
 
 /**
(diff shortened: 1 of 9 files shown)


stevyhacker:
Is it worth it for such a small saving? I assume it's low risk but still wouldn't like to compromise for small gains.

steven2308:
> Is it worth it for such a small saving? I assume it's low risk but still wouldn't like to compromise for small gains.

Not sure TBH. But I do I feel it's such a low impact.

ThunderDeliverer:
Hm, I'd keep this as a possible reduction for now, but wouldn't remove it at this point.. 🤔 

I think reentrancy guard is valuable here no matter the impact, so I wouldn't remove it if it isn't absolutely necessary.

steven2308:
> Hm, I'd keep this as a possible reduction for now, but wouldn't remove it at this point.. thinking
> 
> I think reentrancy guard is valuable here no matter the impact, so I wouldn't remove it if it isn't absolutely necessary.

Let's keep this open for a while. I'm working on a sort of minified version for equippable which should help us at least fit the factories.

steven2308:
We discarded this

PR #177, opened 2023-03-03
Minified equippable tests

steven2308:
# Description

This was supposed to include [this commit](https://github.com/rmrk-team/evm/commit/0a46876ad03a568bdc3fe3cda0f2bfbafe976d10) where I add a "minified" equippable core implementation. It simply merges the implementations of MA, Nestable and Equippable and removes intermediate steps and no longer necessary overrides. The code is mostly duplicated but I think it's a fair price at this point that the code is so stable. It does not include any extension or library code since after testing each it proved to not help with the size at all, in some cases it even increased it.

The ready to use implementations for equippable now inherit from this version, which saves 0.184 of size using 200 runs.
This PR includes a new mock and test suit for it, going trough all behaviors for MA, Nestable, Equippable and 721. This tests also run for each lego already but not including them again messes up with coverage and would be dangerous since there could be mistakes when merging the main 3 legos.

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

### DIFF
--- a/test/minifiedEquippable.ts
+++ b/test/minifiedEquippable.ts
@@ -65,7 +65,7 @@ async function partsFixture() {
 }
 
 async function slotsFixture() {
-  const catalogSymbol = 'SSB';
+  const catalogSymbol = 'SSC';
   const catalogType = 'mixed';
 
   const soldierName = 'SnakeSoldier';
--- a/test/minifiedEquippable.ts
+++ b/test/minifiedEquippable.ts
@@ -1,22 +1,30 @@
+// Note: This is just a copy of the equippable test suit with the additional tests
+// for ERC721 and nestable behavior, but using MinifiedEquippable instead.
+
 import { Contract } from 'ethers';
 import { ethers } from 'hardhat';
 import { expect } from 'chai';
 import { loadFixture } from '@nomicfoundation/hardhat-network-helpers';
 import {
+  addAssetEntryEquippablesFromMock,
   addAssetToToken,
   mintFromMock,
   nestMintFromMock,
-  addAssetEntryEquippablesFromMock,
+  nestTransfer,
+  parentChildFixtureWithArgs,
+  transfer,
 } from './utils';
 import { setupContextForParts } from './setup/equippableParts';
 import { setupContextForSlots } from './setup/equippableSlots';
 import shouldBehaveLikeEquippableAssets from './behavior/equippableAssets';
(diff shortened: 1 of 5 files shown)


test/equippable.ts
```diff
@@ -132,24 +132,11 @@ async function assetsFixture() {
...
-
-  return { equip, renderUtils };
-}
-
 // --------------- END FIXTURES -----------------------
 
 // --------------- EQUIPPABLE BEHAVIOR -----------------------
 
-describe('EquippableMock with Parts', async () => {
+describe('MinifiedEquippableMock with Parts', async () => {
```
ThunderDeliverer, line 139: I think this shouldn't have been modified, right? Same for other instances in this file.. 🤔

test/minifiedEquippable.ts
```diff
@@ -0,0 +1,250 @@
...
+import {
+  RMRKCatalogMock,
+  RMRKMinifiedEquippableMock,
+  RMRKEquipRenderUtils,
+  RMRKMultiAssetRenderUtils,
+} from '../typechain-types';
+// --------------- FIXTURES -----------------------
+
+async function partsFixture() {
+  const catalogSymbol = 'NCB';
```
ThunderDeliverer, line 34: Not really relevant, but I suppose the `B` comes from `Base`, right? We could change this to:
```suggestion
  const catalogSymbol = 'NCC';
```

test/equippable.ts
steven2308: Actually yes. I use the name of the mock in the describe to identify it easily when all tests run

test/minifiedEquippable.ts
```diff
@@ -0,0 +1,250 @@
...
+  // View
+  const view = <RMRKEquipRenderUtils>await viewFactory.deploy();
+  await view.deployed();
+
+  await setupContextForParts(catalog, neon, neon, mask, mask, mintFromMock, nestMintFromMock);
+  return { catalog, neon, mask, view };
+}
+
+async function slotsFixture() {
+  const catalogSymbol = 'SSB';
```
ThunderDeliverer, line 68: This is my final suggestion:

```suggestion
  const catalogSymbol = 'SSC';
```

ThunderDeliverer (approved):
Just this and then I'm happy! 😁

test/equippable.ts
ThunderDeliverer: Oh, I get it. Ok then 😁

PR #180, opened 2023-03-13
Update Emote Tracker in Lue of Redesigning the EIP

ThunderDeliverer:
# Description

This PR updates the Emote Tracker in lue of redesigning the EIP. 
It is helpful to add the access to the mapping here, so that it's as similar to the Repository that will be standardized by the EIP as possible (might help with our discovery as well).

# Actions

The getEmotesPerAddress was added, so that the mapping can be accessed.

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier


steven2308 (changes requested):
IDK if I woke up grumpy but I'm not very happy with the changes :sweat_smile: 
1. I know I added this mapping in the first place but I don't like the `getEmotesPerAddress` name at all. Maybe `didEmoterReactOnToken`.
2. Why do we need that exposed anyway? We just use it to avoid double emoting and allow unemoting.
3. If we're adding anyway, let's return a bool. The int trick is only useful for storing, but outside we can use bool for better ux.

ThunderDeliverer:
Did you sleep for tool long again? 😂 

1. Sure thing. I just accepted the name and continued from there. How about `_emoterReactions`?
2. I imagine this could be quite useful if you want to present a quick emote menu and highlight the already emoted emotes. Similar to what you can do in Discord (see the attached image).
3. So you'd still store an uint, but return a bool? Sure thing, this would be easier to explain.

image

steven2308:
> Did you sleep for too long again? joy
> 
> 1. Sure thing. I just accepted the name and continued from there. How about `_emoterReactions`?
For the inner fine, but for the external I prefer something more like a question, since it returns a bool.
> 2. I imagine this could be quite useful if you want to present a quick emote menu and highlight the already emoted emotes. Similar to what you can do in Discord (see the attached image).
Makes total sense. I'm glad you caught it
> 3. So you'd still store an uint, but return a bool? Sure thing, this would be easier to explain.
Yep!

ThunderDeliverer:
How about `hasEmoterUsedEmote`? For the external one 🤔

steven2308:
> How about `hasEmoterUsedEmote`? For the external one thinking

Sounds great to me.. Let's rename the internal one too then :smile:

ThunderDeliverer:
Ok, I've applied the renames. 🦾  @steven2308 would you take another look?

PR #186, opened 2023-04-03
Add token properties repository

ThunderDeliverer:
# Description

Token properties repository can be used to store and retrieve token properties on chain. It requires a very specific access control and both were added in order to provide as unified experience as possible.

# Actions

- Added token properties repository
- Added token properties repository access control
- Added errors to errors library
- Added tests

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

---

## PR-Codex overview
This PR adds new functionality to the RMRK token properties extension and fixes some errors. It includes adding a new interface for the token properties repository, updating the access control settings, and adding new error messages.

### Detailed summary
- Added `IRMRKTokenPropertiesRepository` interface for the token properties repository
- Updated access control settings for specific parameters
- Added new error messages for various scenarios

> The following files were skipped due to too many changes: `contracts/RMRK/extension/tokenProperties/IRMRKTokenPropertiesRepository.sol`, `docs/RMRK/extension/tokenProperties/IRMRKTokenPropertiesRepository.md`, `contracts/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol`, `yarn.lock`, `docs/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.md`, `test/extensions/tokenPropertiesRepository.ts`


contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,157 @@
...
+    enum AccessType {
+        Issuer,
+        Collaborator,
+        IssuerOrCollaborator,
+        TokenOwner,
+        SpecificAddress
+    }
+
+    /**
+     * @notice Used to noitfy listeners that the access control settings for a specific parameter have been updated.
```
steven2308, line 45:
```suggestion
     * @notice Used to notify listeners that the access control settings for a specific parameter have been updated.
```

contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,157 @@
...
+     * @param collection Address of the collection.
+     * @param parameterType The type of parameter for which the access control settings have been updated.
+     * @param parameterId The ID of the parameter for which the access control settings have been updated.
```
steven2308, line 48: What's this Id? Is it some mapping from the property name? If that's the case, can't we use the name directly?

contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,157 @@
...
+     * @param collection The address of the collection that will use the RMRK token properties repository.
+     * @param issuer The address of the issuer of the collection.
+     * @param useOwnable The boolean value to indicate if the collection implements the Ownable interface and whether it
+     *  should be used to validate that the caller is the issuer (`true`) or to use the manually set issuer address
+     *  (`false`).
+     */
+    function registerAccessControl(
+        address collection,
+        address issuer,
+        bool useOwnable
```
steven2308, line 78: I would simply not support anything not using ownable. The alternative way, if I understand correctly, would allow anyone to claim ownership of some collection's properties

contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,157 @@
...
+     */
+    function manageAccessControl(
+        address collection,
+        ParameterType parameterType,
+        uint256 parameterId,
+        AccessType accessType,
+        address[] memory collaboratorAddresses,
+        bool[] memory collaboratorAddressAccess,
+        address[] memory specificAddresses,
+        bool[] memory specificAddressAccess
```
steven2308, line 122: I don't understand this interface. 
First, what's the need for the bool array? Is it to enable/disable at the same time? It's weird anyway. I expected collaborators to be set by the owner per collection, no need to set them for every property.
Regarding the specific addresses, it would be much simpler if it was just one. So no need to use 2 lists, just a single address which will replace whatever there is for the collection and parameter (which I think should be gated by name, and not id, the id should be an inner thing the users don't have to care about).
If we want to still have a list, let's have only that, no need for the boolean list. Everytime we set it it can simply replace the list, the bad thing is that it would imply a for loop when setting a property to check if the caller is in that list (for the case access type is specific address)

contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,157 @@
...
+     * @param parameterId ID of the parameter being checked.
+     * @return Boolean value indicating if the address is a collaborator of the given collection's parameter (`true`) or
+     *  not (`false`).
+     */
+    function isCollaborator(
+        address collaborator,
+        address collection,
+        ParameterType parameterType,
+        uint256 parameterId
+    ) external view returns (bool);
```
steven2308, line 139: As stated previously:
1. I think collaborators should be independent from properties
2.I don't see the need for parameter type here, or anywhere in this contract. You shouldn't have multiple properties with different type and the same name, I would add it as a restriction.
3. Parameter Id is harder to use than simply a name

contracts/RMRK/extension/tokenProperties/IRMRKTokenPropertiesRepository.sol
```diff
@@ -0,0 +1,147 @@
...
+     * @param tokenId The token ID
+     * @param key The key of the property
+     * @param value The new value of the property
+     */
+    event BytesPropertyUpdated(
+        address indexed collection,
+        uint256 indexed tokenId,
+        string key,
+        bytes value
+    );
```
steven2308, line 81: I think these events are perfect

contracts/RMRK/extension/tokenProperties/IRMRKTokenPropertiesRepository.sol
```diff
@@ -0,0 +1,147 @@
...
+     * @param tokenId The token ID
+     * @param key The key of the property
+     * @return The value of the bytes property
+     */
+    function getBytesTokenProperty(
+        address collection,
+        uint256 tokenId,
+        string memory key
+    ) external view returns (bytes memory);
+}
```
steven2308, line 147: Getters also look perfect.

contracts/RMRK/extension/tokenProperties/RMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,251 @@
...
+
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+
+import "./IRMRKPropertiesAccessControl.sol";
+
+contract RMRKPropertiesAccessControl is IRMRKPropertiesAccessControl {
+    // mapping(address collection => mapping (ParameterType => mapping (uint256 parameterId => AccessControl))) private _accessControl;
+    mapping(address => mapping(ParameterType => mapping(uint256 => AccessControl)))
```
steven2308, line 11: No need for ParameterType here IMO.

contracts/RMRK/extension/tokenProperties/RMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,251 @@
...
+        private _accessControl;
+    // mapping(address collection => IssuerSetting issuerSetting) private _issuerSettings;
```
steven2308, line 13: I suspect you're commenting this to be able to run coverage. I solved it on markeplace repo with this: https://github.com/rmrk-team/evm-marketplace/commit/613fd558cf31268b52835fcf625efefeaa955312

contracts/RMRK/extension/tokenProperties/RMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,251 @@
...
+        }
+        _;
+    }
+
+    /**
+     * @notice Modifier to check if the caller is the issuer of the collection.
+     * @param collection Address of the collection.
+     */
+    modifier onlyIssuer(address collection) {
+        if (_issuerSettings[collection].useOwnable) {
```
steven2308, line 182: I think it's a laudable effort but to dangerous to keep. Let's only support ownable IMO

contracts/RMRK/extension/tokenProperties/RMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,251 @@
...
+     * @param parameterType The type of the parameter.
+     * @param parameterId The ID of the parameter.
+     * @param tokenId The ID of the token.
+     */
+    function _onlyAuthorizedCaller(
+        address collection,
+        ParameterType parameterType,
+        uint256 parameterId,
+        uint256 tokenId
+    ) private view {
```
steven2308, line 212: This looks a lot like what I expected. Just without the parameter type.
I see here why you're using Id on the access control, since you make them as independent as possible. TBH I would just merge them so I could use getIdForKey inside the access control configuration and not have the user figure out what the id even means in that context.

contracts/RMRK/extension/tokenProperties/RMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,251 @@
...
+        ) {
+            revert RMRKNotCollectionIssuerOrCollaborator();
+        } else if (
+            accessControl.accessType == AccessType.TokenOwner &&
+            IERC721(collection).ownerOf(tokenId) != msg.sender
+        ) {
+            revert RMRKNotTokenOwner();
+        } else if (
+            accessControl.accessType == AccessType.SpecificAddress &&
+            !accessControl.specificAddresses[msg.sender]
```
steven2308, line 246: Also kind of see why you're using the mapping to config the access control, but I still really dislike the bool[] + address[] interface. This might be cheaper though TBH, but more complex UX. :thinking:

contracts/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol
```diff
@@ -0,0 +1,286 @@
...
+            ParameterType.ADDRESS,
+            _getIdForKey(collection, key),
+            tokenId
+        )
+    {
+        _addressValues[collection][tokenId][
+            _getIdForKey(collection, key)
+        ] = value;
+        emit AddressPropertyUpdated(collection, tokenId, key, value);
+    }
```
steven2308, line 227: I think all this setters are prefect with one caveat. Since this is not an extensions, these should already be the external versions.

contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
steven2308: I later realized it was the id from keyToId we use internally in token properties. I don't think the user needs to know about it.

contracts/mocks/extensions/tokenProperties/RMRKTokenPropertiesRepositoryMock.sol
```diff
@@ -0,0 +1,87 @@
...
+pragma solidity ^0.8.18;
+
+import "../../../RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol";
+
+/**
+ * @title RMRKTokenPropertiesRepositoryMock
+ * @author RMRK team
+ * @notice Smart contract of the RMRK Token properties module.
+ */
+contract RMRKTokenPropertiesRepositoryMock is RMRKTokenPropertiesRepository {
```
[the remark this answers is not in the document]
ThunderDeliverer: You are right, we don't need internal functions either. I'll update accordingly 🦾

contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
```diff
@@ -0,0 +1,157 @@
...
+interface IRMRKPropertiesAccessControl {
+    /**
+     * @notice A list of supported parameter types.
+     * @return The `string` type.
+     * @return The `address` type.
+     * @return The `bytes` type.
+     * @return The `uint` type.
+     * @return The `bool` type.
+     */
+    enum ParameterType {
```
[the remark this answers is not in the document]
ThunderDeliverer: This was designed with the assumption that the same key can be used for different types, but upon reading your comments I realised this doesn't make sense.. I'll remove this layer of access control 👌

contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
ThunderDeliverer: Yeah, I think this is better than using keys as it is a fixed size variable. We need to figure out how to package this for the proposal.. 🤔

I'd leave the documentation as it explains how we are approaching the access control if someone with sufficient knowledge wants to verify the operation

contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
ThunderDeliverer: I'd leave it. I think it's beneficial to have the ability for all the collections to use the repository. Yes, anyone can utilise the repository to claim the issuer role if the collection doesn't implement Ownable, but I don't see this as a big issue.

1. Only using Ownable bars any collection that doesn't have `owner()` implemented.
2. Allowing to claim collections that don't have `owner()` is inclusive of all collections. There is a risk that a 3rd party could claim it (this would mean that the legitimate issuer can't use it, which is essentially the same as option 1), but there is also the ability to use this if the collection uses some other kind of access control or if it is self-owned.

I think option 2 allows for more inclusion.. 🤔

contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
ThunderDeliverer: Yeah, the idea was to allow to add and remove any number of collaborators at a time and leave the states of already set ones alone if needed.
Having collaborators defined on a collection level makes sense, I agree and I'll modify this. 👌 
Ok, I don't like the idea of the for loop, so I think a specific address on a per-key basis seems best. 🤔

contracts/RMRK/extension/tokenProperties/IRMRKPropertiesAccessControl.sol
ThunderDeliverer: We agree on the 1st point. 👌 
We agree on the 2nd point. 👌 
I still think we should keep it. The user never needs to pass the ID for anything else than management. We could expose a method to get ID for key, so that this process gets easier as well 🤔

contracts/RMRK/extension/tokenProperties/RMRKPropertiesAccessControl.sol
ThunderDeliverer: Agree, it will be removed from all of the access control 👌

contracts/RMRK/extension/tokenProperties/RMRKPropertiesAccessControl.sol
ThunderDeliverer: Why do you think this is dangerous?

contracts/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol
ThunderDeliverer: 🦾

contracts/RMRK/extension/tokenProperties/RMRKPropertiesAccessControl.sol
ThunderDeliverer: Amazing!

image

contracts/RMRK/extension/tokenProperties/RMRKPropertiesAccessControl.sol
steven2308: Allowing anyone to claim a way of ownership can lead to ransoms or phishing attempts. You could more easily convince holders that you are the rightful issuer by modifying the attributes of  a well known centralized attributes repo.

ThunderDeliverer: While this diminishes the list of potential collection users, I agree with you on the risks. It simplifies things from our end and makes the smart contract more reliable 🤔

contracts/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol
```diff
@@ -0,0 +1,471 @@
...
+    mapping(address => mapping(uint256 => AccessType))
+        private _parameterAccessType;
+    mapping(address => mapping(uint256 => address))
+        private _parameterSpecificAddress;
+    mapping(address => IssuerSetting) private _issuerSettings;
+    mapping(address => mapping(address => bool)) private _collaborators;
+
+    // For keys, we use a mapping from strings to IDs.
+    // The purpose is to store unique string keys only once, since they are more expensive,
+    mapping(address => mapping(string => uint256)) private _keysToIds;
```
steven2308, line 27: There's no need to do it by collection. We can make it general and save some gas. If multiple collections use the same Key (which I suspect they will), only the first one creates the key, the rest are cheaper.

contracts/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol
```diff
@@ -0,0 +1,471 @@
...
+        address issuer;
+    }
+
+    /**
+     * @inheritdoc IRMRKTokenPropertiesRepository
+     */
+    function registerAccessControl(
+        address collection,
+        address issuer,
+        bool useOwnable
```
steven2308, line 60: I thought we agreed on removing this :grimacing:  Besides risky, it makes the code more complex on a few places
https://github.com/rmrk-team/evm/pull/186#discussion_r1161619697

contracts/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol
```diff
@@ -0,0 +1,471 @@
...
+     */
+    function manageCollaborators(
+        address collection,
+        address[] memory collaboratorAddresses,
+        bool[] memory collaboratorAddressAccess
+    ) external onlyRegisteredCollection(collection) onlyIssuer(collection) {
+        if (collaboratorAddresses.length != collaboratorAddressAccess.length) {
+            revert RMRKCollaboratorArraysNotEqualLength();
+        }
+        for (uint256 i = 0; i < collaboratorAddresses.length; i++) {
```
steven2308, line 117: For gas savings, no need to do `i = 0` since 0 is the default. Also, `unchecked { ++i; }` is cheaper for increase the counter

steven2308, line 117: For gas savings, no need to do` = 0`, since 0 is the default. Also `unchecked { ++i; }` is cheaper.

contracts/RMRK/extension/tokenProperties/IRMRKTokenPropertiesRepository.sol
```diff
@@ -0,0 +1,288 @@
...
+     * @param issuer The address of the issuer of the collection.
+     * @param useOwnable The boolean value to indicate if the collection implements the Ownable interface and whether it
+     *  should be used to validate that the caller is the issuer (`true`) or to use the manually set issuer address
+     *  (`false`).
+     */
+    function registerAccessControl(
+        address collection,
+        address issuer,
+        bool useOwnable
+    ) external;
```
steven2308, line 155: Removing support for non ownable implementer also removes the need for this I think. The issuer can just start managing access control for properties directly.

contracts/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol
ThunderDeliverer: Yeah, I considered this, but then started thinking about how this would behave through decades and decided to leave it 🤔 
It's an insanely huge number for the possible IDs, but still..
I can remove it if you think it's future-proof without doing it per collection 👌

contracts/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol
ThunderDeliverer: I thought we agreed only on removing the ability to use the collection that doesn't have ownable implemented 🤔

steven2308 (changes requested):
Not sure if I got it wrong that we had agreed on removing non ownable support. That would simplify a lot the code.
Besides that, just a couple of suggestions.

contracts/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol
steven2308: It's still 1.15e77 IDs :sweat_smile:  but seen that way, not strongly opposed to change it. We can also see what community thinks.

ThunderDeliverer: This also means that we don't need to keep the `_totalProperties` value per-contract.. 🤔

contracts/RMRK/extension/tokenProperties/RMRKTokenPropertiesRepository.sol
steven2308: All we care is that it has owner() method, even if they don't claim support the interface.
If they have it good, owner() we can register properties. If not, sorry for them :man_shrugging:

ThunderDeliverer: Yeah, this implementation does just that. You don't need to use ownable when using it, but you need the `owner` method in order to register access control.

steven2308 (approved):

steven2308 (approved):
LFG!

PR #198, opened 2023-05-06
Add nestable utils

ThunderDeliverer:
# Description

We need a utility to validate that the token actually is the parent of the child tokens that it reports as their child tokens.

# Actions

RMRKNestableUtils were added. Their current functionality only covers validating whether child tokens belong to the parent token or not, but they should expand over time.

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

---

## PR-Codex overview
This PR adds two new errors and two new functions to the RMRK library. The functions are used to validate whether a child token is owned by a given parent token. The changes are as follows:

### Detailed summary
- Added `RMRKMismachedArrayLength` error
- Added `validateChildOf` function to validate if a child token is owned by a parent token
- Added `validateChildrenOf` function to validate if multiple child tokens are owned by a parent token
- Updated documentation to reflect changes


ThunderDeliverer:
I only managed to add the util, and I'll add the tests tomorrow, as it's already pretty late.

ThunderDeliverer:
I'm not sure how to link ClickUp, but this is me trying 😂  #866a80pd4

contracts/RMRK/utils/RMRKNestableUtils.sol
```diff
@@ -0,0 +1,122 @@
...
+        if (
+            !IERC165(childAddress).supportsInterface(type(IERC6059).interfaceId)
+        ) {
+            return false;
+        }
+
+        (address directOwner, uint256 ownerId, ) = IERC6059(childAddress)
+            .directOwnerOf(childId);
+
+        if (directOwner == parentAddress && ownerId == parentId) {
```
steven2308, line 39: A detail, but you could directly return the contents of this `if`

contracts/RMRK/utils/RMRKNestableUtils.sol
```diff
@@ -0,0 +1,122 @@
...
+        address[] memory tmpInvalidChildAddresses = new address[](
+            childAddresses.length
+        );
+        uint256[] memory tmpInvalidChildIds = new uint256[](
+            childAddresses.length
+        );
+        bool isValid = true;
+
+        for (uint256 i; i < childAddresses.length; ) {
+            if (
```
steven2308, line 78: Any reason not to reuse the single child version here?

contracts/RMRK/utils/RMRKNestableUtils.sol
```diff
@@ -0,0 +1,122 @@
...
+
+            delete directOwner;
+            delete ownerId;
+
+            unchecked {
+                ++i;
+            }
+        }
+
+        address[] memory invalidChildAddresses = new address[](
```
steven2308, line 104: What if you simple returned an array of bools indicating for each child, in the same order, whether it was valid or not? 
Would be much cheaper and should be equally useful to the front.

steven2308 (approved):
Very detailed tests, nice!

Added some suggestions.

contracts/RMRK/utils/RMRKNestableUtils.sol
ThunderDeliverer: Right 😅 🙈

contracts/RMRK/utils/RMRKNestableUtils.sol
ThunderDeliverer: This should be cheaper when called and since we have a lot of space for this util, I thought it sensible to do it this way.

steven2308: I don't think in this case gas savings are a greater good than maintainability. Also, it is a view. I know it could be used from a method altering the state, but that would be rare.

steven2308 (changes requested):
Somehow I missed this before. But there is already RMRKNestableRenderUtils. All tests are in tests/renderUtils.ts.
Could you move code and tests to the existing ones?

ThunderDeliverer:
> Somehow I missed this before. But there is already RMRKNestableRenderUtils. All tests are in tests/renderUtils.ts. Could you move code and tests to the existing ones?

This is purely a check for operation, this is why I assumed that the task called for the separate util form RMRKNestableRenderUtil. This doesn't help with rendering, just with validation 🤔

steven2308:
Maybe we should simply rename them all to utils. I think it makes sense to have a single contract for them.

steven2308 (approved):
I think they should be merged, under whatever name you consider. But it is more useful to have them all in a single place.
I'll leave it up to you

steven2308 (commented):
I see you merged the contracts code, and deleted the tests, but I don't see the tests on the renderUtils file. Maybe you forgot to commit?

ThunderDeliverer:
> I see you merged the contracts code, and deleted the tests, but I don't see the tests on the renderUtils file. Maybe you forgot to commit?

Seems that way, yeah. It should be ok now..

PR #211, opened 2023-06-01
ADds withdrawRaisedERC20 method on all ERC20Pay implementations.

steven2308:
# Description

I did not add it on minting utils directly since it would unnecesarily increase the size for all implementations. 
The method expects the ERC20 token address, it could be retrieved from contract but it is more expensive in gas and size, and this way you could even recovered other ERC20s sent by mistake to the contract.

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

---

## PR-Codex overview
This PR adds a `withdrawRaisedERC20` function to multiple ERC20 payment implementation contracts and updates the `utils.ts` and `IERC20.sol` files to support it.

### Detailed summary
- Added `withdrawRaisedERC20` function to `RMRKMultiAssetImplErc20Pay.sol`, `RMRKNestableImplErc20Pay.sol`, `RMRKEquippableImplErc20Pay.sol`, and `RMRKNestableMultiAssetImplErc20Pay.sol`
- Updated `IERC20.sol` file to include `transfer` function documentation
- Updated `test/utils.ts` file to import `BigNumber` from `ethers`
- Added tests for `withdrawRaisedERC20` function in `lazyMintErc20Pay.ts` test file

### DIFF
--- a/contracts/implementations/erc20Pay/RMRKEquippableImplErc20Pay.sol
+++ b/contracts/implementations/erc20Pay/RMRKEquippableImplErc20Pay.sol
@@ -103,4 +103,19 @@ contract RMRKEquippableImplErc20Pay is
     function _charge(uint256 value) internal virtual override {
         _chargeFromToken(msg.sender, address(this), value);
     }
+
+    /**
+     * @notice Used to withdraw the minting proceedings to a specified address.
+     * @dev This function can only be called by the owner.
+     * @param erc20 Address of the ERC20 token to withdraw
+     * @param to Address to receive the given amount of minting proceedings
+     * @param amount The amount to withdraw
+     */
+    function withdrawRaisedERC20(
+        address erc20,
+        address to,
+        uint256 amount
+    ) external onlyOwner {
+        IERC20(erc20).transfer(to, amount);
+    }
 }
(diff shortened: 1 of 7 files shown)


contracts/implementations/erc20Pay/RMRKEquippableImplErc20Pay.sol
```diff
@@ -103,4 +103,19 @@ contract RMRKEquippableImplErc20Pay is
...
+     * @dev This function can only be called by the owner.
+     * @param erc20 Address of the ERC20 token to withdraw
+     * @param to Address to receive the given amount of minting proceedings
+     * @param amount The amount to withdraw
+     */
+    function withdrawRaisedERC20(
+        address erc20,
+        address to,
+        uint256 amount
+    ) external onlyOwner {
```
ThunderDeliverer, line 118: I was considering whether we should add a reentrancy guard here as it is an external call 🤔

But I think we should be ok.. Even if the `erc20` contract is malicious and attempts to drain another token using `delegatecall`, the `msg.sender` would be this smart contract's address, so it would revert on `onlyOwner` check 🤔

ThunderDeliverer (approved):

contracts/implementations/erc20Pay/RMRKEquippableImplErc20Pay.sol
steven2308: Yeah this feels pretty safe

PR #223, opened 2023-07-19
Added Mumbai to Hardhat config

steven2308:
# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

---

## PR-Codex overview
This PR focuses on adding support for Polygon Mumbai network and its corresponding Etherscan API Key. 

### Detailed summary
- Added `polygonMumbai` network configuration in `hardhat.config.ts`.
- Added `url`, `chainId`, `accounts`, and `gasPrice` properties for `polygonMumbai` network.
- Added `polygonMumbai` Etherscan API Key in the `config` object.

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -60,7 +60,7 @@ const config: HardhatUserConfig = {
       chainId: 11155111,
       accounts: process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
     },
-    mumbai: {
+    polygonMumbai: {
       url: process.env.MUMBAI_URL || '',
       chainId: 80001,
       accounts: process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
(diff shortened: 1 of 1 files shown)


hardhat.config.ts
```diff
@@ -60,6 +60,12 @@ const config: HardhatUserConfig = {
       chainId: 11155111,
       accounts: process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
     },
+    mumbai: {
```
steven2308, line 63:
```suggestion
    polygonMumbai: {
```
This should match with etherscan keys. And I think it will be clearer.  Or let's set mumbai on etherscan keys.

ThunderDeliverer: For some reason, setting etherscan kay to `mumbai` fails as it requires `polygonMumbai` there. But we could try applying your suggestion and see if everything works as expected.

This is the only network that I've seen such behavior, but the mismatched named keys work for some reason.

steven2308: It tried locally in both EVM and marketplace with `polygonMumbai` in both in worked well so I'll apply it :smile:

PR #233, opened 2023-10-20
Royalties splitter

steven2308:
# Description

Adds RoyaltiesSplitter contract.
Beneficiaries and shares are set on creation and cannot be further changed.
Native tokens received are immediately distributed.
ERC20 tokens do not notify the receiver, so they must be manually claimed. To simplify the logic, the distribute method can be called by anyone and it will distribute to all beneficiaries according to the shares.

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

---

### Detailed summary
- Added `RMRKRoyaltiesSplitter` smart contract for sharing royalties from native or ERC20 payments.
- Added `distributeERC20` function to distribute ERC20 payments to beneficiaries.
- Added `getBenefiariesAndShares` function to get the list of beneficiaries and their shares.
- Added `ERCPaymentDistributed` and `NativePaymentDistributed` events.
- Added error types `InvalidInput`, `FailedToSend`, and `OnlyBeneficiary`.
- Added tests for the `RMRKRoyaltiesSplitter` contract.

### DIFF
--- a/contracts/implementations/utils/RMRKRoyaltiesSplitter.sol
+++ b/contracts/implementations/utils/RMRKRoyaltiesSplitter.sol
@@ -1,0 +1,86 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.21;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+error InvalidInput();
+error FailedToSend();
+
+/**
+ * @title RMRKRoyaltiesSplitter
+ * @author RMRK team
+ * @notice Smart contract of the RMRK Royalties Spliter module.
+ * @dev This smart contract provides a simple way to share royalties from either native or erc20 payments.
+ */
+contract RMRKRoyaltiesSplitter {
+    event NativePaymentDistributed(address indexed sender, uint256 amount);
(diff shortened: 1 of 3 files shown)


Swader:
What if the artists have a falling out, or the collection is sold, and there is a need to change one of the recipients? Or is this a separate tool that is set as the royalty recipient, and then middleman style distributes the funds?

steven2308:
> What if the artists have a falling out, or the collection is sold, and there is a need to change one of the recipients? Or is this a separate tool that is set as the royalty recipient, and then middleman style distributes the funds?

In that case you would simply deploy a new splitter and update the royalties recipient on the collection. The collection owner still holds the keys and could update the recipient any time, so this requires trust on them, unless it's owned by a multisig.

Swader:
Yep, makes sense, all good then.

ilionic (approved):

ilionic:
Would be good to add additional distribution gating mechanics, ie restricting to beneficiaries

steven2308:
> Would be good to add additional distribution gating mechanics, ie restricting to beneficiaries

Now only beneficiaries can init the distribution for ERC20s

PR #267, opened 2024-03-19
Merge after 2.5.3.

steven2308:
Update after v2.5.3

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier


contracts/RMRK/extension/tokenHolder/RMRKTokenHolder.sol
```diff
@@ -94,7 +102,14 @@ abstract contract RMRKTokenHolder is IERC7590 {
             amount,
             data
         );
-        IERC20(erc20Contract).transferFrom(msg.sender, address(this), amount);
+        IERC20 erc20 = IERC20(erc20Contract);
+        uint256 initBalance = erc20.balanceOf(address(this));
+        erc20.transferFrom(msg.sender, address(this), amount);
+        uint256 newBalance = erc20.balanceOf(address(this));
+        // Here you can either use the difference as the amount, or revert if the difference is not equal to the amount and you don't want to support transfer fees
```
Yuripetusko, line 109: to me this comment is not very clear tbh

Yuripetusko (approved):

contracts/RMRK/extension/tokenHolder/RMRKTokenHolder.sol
steven2308: Turns out there are ERC20 tokens with "Tax". So you call `token.transfer(10)` but only 9.9 are sent.
Initially, the token holder was assuming that the transferred value was received, so the balances of the NFT would be 10, but the contract would actually only have 9.9 in this case.
To fix this there are 2 options:
1. If the balance of the contract did not increase by the exact amount, then it reverts. (Safest, but does not support tokens with taxes).
2. Instead of increasing the token's balance to what was supposedly transfer, only increase it by the actually increased balance.
I implemented 1.

Yuripetusko: thanks, but I meant more about the comment :D maybe if I didn't get it, others won't too, so we should improve the comment

PR #269, opened 2024-04-19
Implements BulkWriter.transferAndEquip.

steven2308:
# Description

The NFT to transfer can be equipped, nested or free. The destination must be an NFT.
The NFT can be equipped into the destination.
If destination slot is occupied, it will be unequipped. 
All these operations happen under a single transactions, given the right approvals.

# Checklist

- [x] Verified code additions
- [x] Regenerated docs
- [x] Ran prettier

### DIFF
--- a/contracts/RMRK/utils/RMRKBulkWriter.sol
+++ b/contracts/RMRK/utils/RMRKBulkWriter.sol
@@ -27,6 +27,14 @@ contract RMRKBulkWriter is Context {
         uint64 slotPartId;
     }
 
+    struct ParentData {
+        address parentCollection;
+        uint256 parentTokenId;
+        uint256 childIndexInParent;
+        uint64 parentEquippedAssetId;
+        uint64 parentEquippedSlotPartId;
+    }
+
     /**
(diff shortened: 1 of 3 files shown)


contracts/RMRK/utils/RMRKBulkWriter.sol
```diff
@@ -216,6 +224,126 @@ contract RMRKBulkWriter is Context {
...
+        _acceptChildIfNecessary(
+            destinationCollection,
+            destinationTokenId,
+            collection,
+            tokenId
+        );
+
+        if (equipData.assetId == 0) return;
+
+        _unequipSlotIfNecessary(
```
Yuripetusko, line 295: wondering if there's any benefit in re-using replaceEquipement function from this contract here instead?

Yuripetusko (approved):

contracts/RMRK/utils/RMRKBulkWriter.sol
steven2308: Sound great, will give it a try.

